PR #206, opened 2023-06-15
move span-context to a separate file

yousif-bugsnag:
## Goal

Adds a new `SpanContextStorage` interface and a default implementation in core using a backing array for the context stack

## Design

```ts
export interface SpanContextStorage extends Iterable<SpanContext> {
  push: (context: SpanContext) => void
  pop:  (context: SpanContext) => void
  readonly current: SpanContext | undefined
}
```

- `current` returns the current `SpanContext` at the top of the stack  if it `isValid()`, or undefined if the stack is empty
 - `push` adds to the top of the stack and makes context the new current if it `isValid()`
- `pop` will remove the top of the stack if current === context
	- will be called by the `SpanFactory` when a Span is ended
- `pop` and `current` will both remove any invalid SpanContexts from the top of the stack before returning

`SpanContextStorage` is also an Iterable<SpanContext> - this is to allow the entire stack to be scanned by the SpanFactory (and possibly other span creators) as some attributes will need to be determined by what is already present in the current stack


packages/core/lib/span-context.ts
```diff
@@ -0,0 +1,89 @@
...
+export class DefaultSpanContextStorage implements SpanContextStorage {
+  private readonly contextStack: SpanContext[]
+
+  constructor (contextStack: SpanContext[] = []) {
+    this.contextStack = contextStack
+  }
+
+  [Symbol.iterator] (): SpanContextIterator {
+    return new SpanContextIterator(this.contextStack)
+  }
```
imjoehaines, line 58: I believe you can replace the whole `SpanContextIterator` with [a `yield*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*):

```suggestion
  *[Symbol.iterator] (): Iterator<SpanContext> {
    return yield* this.contextStack
  }
```

yousif-bugsnag: Oh nice! TIL!

I've removed that iterator class and switched around the logic for `push`/`pop`/`current` so that we're adding contexts to the start of the array and then the iterator iterates in the correct order

imjoehaines: Ah I forgot this was in reverse order!

We don't want to have to `shift`/`unshift` the backing array since that's a lot slower than `push`/`pop`, so an implementation like this would be better:

```ts
class DefaultSpanContextStorage {
  *[Symbol.iterator] (): Iterator<SpanContext> {
    for (let i = this.contextStack.length - 1; i >= 0; --i) {
      yield this.contextStack[i]
    }
  }
}
```

imjoehaines (approved):

PR #208, opened 2023-06-15
Apply a 100ms cooldown period in onSettle

imjoehaines:
## Goal

We now wait 100ms in case the settlers unsettle immediately after calling 'onSettle'

This ensures we aren't settling over-eagerly when the page is _about_ to change (or make network requests) but hasn't actually done so at the point 'onSettle' is called

For example, previously this would have settled immediately but now it will not as the DOM is mutated in the 100ms "cooldown period":

```ts
const span = BugsnagPerformance.startSpan('a')

onSettle(() => { span.end() })

document.body.innerHTML = ':)'
```

### DIFF
--- a/packages/platforms/browser/lib/on-settle/index.ts
+++ b/packages/platforms/browser/lib/on-settle/index.ts
@@ -36,7 +36,7 @@ export default function createOnSettle (
     xhrRequestSettler
   ])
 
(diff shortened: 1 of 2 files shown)


yousif-bugsnag (approved):
took me a minute to get my head around it but got there in the end! LGTM

PR #210, opened 2023-06-15
Make `SpanContextStorage` a client option and add `currentSpanContext` to client

yousif-bugsnag:
## Goal

Adds a new `spanContextStorage` client option for platforms to provide their own implementation. There are currently no platforms-specific implementations, so this is optional - if no implementation is provided then `DefaultSpanContextStorage` is used.

Also adds `currentSpanContext` as a new readonly property to the client to allow users to retrieve the current context and use it manually if they need to (for example starting multiple spans under the same parent context)

### DIFF
--- a/packages/core/lib/core.ts
+++ b/packages/core/lib/core.ts
@@ -13,10 +13,12 @@ import { BufferingProcessor, type Processor } from './processor'
 import { InMemoryQueue } from './retry-queue'
 import Sampler from './sampler'
 import { SpanFactory, type Span, type SpanOptions } from './span'
+import { type SpanContext, type SpanContextStorage, DefaultSpanContextStorage } from './span-context'
 
 export interface BugsnagPerformance<C extends Configuration> {
   start: (config: C | string) => void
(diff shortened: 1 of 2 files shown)


packages/core/lib/core.ts
```diff
@@ -84,6 +88,9 @@ export function createClient<S extends CoreSchema, C extends Configuration> (opt
     startSpan: (name, spanOptions?: SpanOptions) => {
       const span = spanFactory.startSpan(name, spanOptions)
       return spanFactory.toPublicApi(span)
+    },
+    get currentSpanContext () {
+      return spanContextStorage.current
```
imjoehaines, line 93: It'd be good to add some tests for this. Should be possible by passing a storage as an `option` and pushing to that

yousif-bugsnag: yeah good point - I was going to test this 'for real' that after implementing `makeCurrentContext`, but it's good to have a test that uses the storage instance provided in client options as well

imjoehaines (approved):

gingerbenw (approved):

PR #214, opened 2023-06-19
Add `parentContext` to `SpanOptions` and set parentage for new spans

yousif-bugsnag:
## Goal

This PR adds a new optional `parentContext` option to `SpanOptions`

```ts
interface SpanOptions {
	parentContext?: SpanContext | null
}
```

A new optional `parentSpanId` field has also been added to internal and delivery spans - this is only set if the span has a parent

When a `parentContext` is specified in `SpanOptions`, the created span will become a child of that context. This means that the span's `parentSpanId` will be set to the parent's `id`, and the span's `traceId` will be set to the parent's `traceId`.

If no `parentContext` is specified, by default new spans will become children of `currentSpanContext`. 

if there is no `currentSpanContext`, or `parentContext` is set explicitly to `null`, the created span is a new root span - this means `parentSpanId` is not set and a new `traceId` is generated 

## Testing

Added unit tests for the new `parentContext` option as well as a new mazerunner scenario to test that nested spans are sent with the correct parentage.

### DIFF
--- a/packages/core/tests/span.test.ts
+++ b/packages/core/tests/span.test.ts
@@ -198,15 +198,14 @@ describe('Span', () => {
     })
 
     it('sets traceId and parentSpanId from parentContext if specified', async () => {
-      let count = 0
+      let spanCount = 0
+      let traceCount = 0
       const idGenerator = {
         generate (bits: 64 | 128) {
           if (bits === 64) {
-            count++
-            return `span ID ${count}`
+            return `span ID ${++spanCount}`
           }
-
-          return `trace ID ${count}`
+          return `trace ID ${++traceCount}`
         }
       }
 
@@ -234,15 +233,14 @@ describe('Span', () => {
     })
 
     it('starts a new root span when parentContext is null', async () => {
-      let count = 0
+      let spanCount = 0
+      let traceCount = 0
       const idGenerator = {
         generate (bits: 64 | 128) {
           if (bits === 64) {
-            count++
-            return `span ID ${count}`
+            return `span ID ${++spanCount}`
           }
-
-          return `trace ID ${count}`
+          return `trace ID ${++traceCount}`
         }
       }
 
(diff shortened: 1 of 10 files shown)


packages/core/tests/span.test.ts
```diff
@@ -196,6 +196,123 @@ describe('Span', () => {
       expect(spanContextEquals(spanIsNotContext, client.currentSpanContext)).toBe(false)
       expect(spanContextEquals(spanIsContext, client.currentSpanContext)).toBe(true)
     })
+
+    it('sets traceId and parentSpanId from parentContext if specified', async () => {
+      let count = 0
+      const idGenerator = {
+        generate (bits: 64 | 128) {
+          if (bits === 64) {
+            count++
```
imjoehaines, line 205: If `count` is only incremented here, doesn't this somewhat invalidate the test since all spans would always be in the same trace?

yousif-bugsnag: doh, yes it does- good spot. That's the whole point of having that generator! Updated so that a new span ID and trace ID is generated each time

packages/core/tests/span.test.ts
```diff
@@ -196,6 +196,120 @@ describe('Span', () => {
...
+        }
+      }
+
+      const delivery = new InMemoryDelivery()
+      const client = createTestClient({ idGenerator, deliveryFactory: () => delivery })
+      client.start({ apiKey: VALID_API_KEY })
+
+      // push two spans onto the context stack
+      const span1 = client.startSpan('span 1')
+      const span2 = client.startSpan('span 2')
```
gingerbenw, line 218:
```suggestion
      const span1 = client.startSpan('should become parent')
      const span2 = client.startSpan('should not become parent')
```

test/browser/features/nested-spans.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Nested spans
+
+  Scenario: Spans are nested under the correct parent context
+    Given I navigate to the test URL "/nested-spans"
+    And I wait for 5 spans
+
```
gingerbenw, line 6: comment me please!

PR #217, opened 2023-06-20
Add `device.id` as a resource attribute

imjoehaines:
## Goal

Add `device.id` as a resource attribute using `@bugsnag/cuid` and the `bugsnag-anonymous-id` key so the IDs are shared with the notifier: https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-browser-device/device.js

This required changing `ResourceAttributeSource` to be asynchronous as we need to fetch the device ID from persistence at least once but can't until we've read the `generateAnonymousId` config option

### DIFF
--- a/packages/platforms/browser/lib/persistence.ts
+++ b/packages/platforms/browser/lib/persistence.ts
@@ -28,6 +28,53 @@ function makeBrowserPersistence (window: WindowWithLocalStorage): Persistence {
   return new InMemoryPersistence()
 }
 
+// NOTE: this should be kept in sync with the notifier
+// https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-browser-device/device.js
+function isDeviceId (raw: string): boolean {
+  // make sure the persisted value looks like a valid cuid
+  return /^c[a-z0-9]{20,32}$/.test(raw)
+}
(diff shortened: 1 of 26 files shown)


yousif-bugsnag (approved):

packages/platforms/browser/lib/persistence.ts
```diff
@@ -28,6 +28,53 @@ function makeBrowserPersistence (window: WindowWithLocalStorage): Persistence {
   return new InMemoryPersistence()
 }
 
+// NOTE: this should be kept in sync with the notifier
+// https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-browser-device/device.js
+function isDeviceId (raw: string): boolean {
+  // make sure the persisted value looks like a valid cuid
+  return /^c[a-z0-9]{20,32}$/.test(raw)
+}
```
gingerbenw, line 36: could we put this in our own fork of cuid, and import it in both packages?

imjoehaines: Yeah, good idea. Also the `.d.ts` file

https://smartbear.atlassian.net/browse/PLAT-10305

PR #218, opened 2023-06-20
Add `isFirstClass` span option controlling the `bugsnag.span.first_class` attribute

imjoehaines:
## Goal

Add `isFirstClass` span option controlling the `bugsnag.span.first_class` attribute

This defaults to `true` for all spans (custom, network, page load & route changes) and can be set to `false` for custom spans only

This necessitated changing the `StartRouteChangeCallback` options parameter to omit `isFirstClass` as route change spans are always first class (this is the same for network and page load spans but users can't supply options for those span types)

### DIFF
--- a/packages/platforms/browser/lib/span-attributes-source.ts
+++ b/packages/platforms/browser/lib/span-attributes-source.ts
@@ -4,7 +4,6 @@ const createSpanAttributesSource = (title: string, url: string): SpanAttributesS
   return () => {
     const spanAttributes = new Map<string, SpanAttribute>()
     spanAttributes.set('bugsnag.span.category', 'custom')
-    spanAttributes.set('bugsnag.span.first_class', true)
     spanAttributes.set('bugsnag.browser.page.url', url)
     spanAttributes.set('bugsnag.browser.page.title', title)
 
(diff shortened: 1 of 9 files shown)


packages/platforms/browser/lib/span-attributes-source.ts
```diff
@@ -4,7 +4,6 @@ const createSpanAttributesSource = (title: string, url: string): SpanAttributesS
   return () => {
     const spanAttributes = new Map<string, SpanAttribute>()
     spanAttributes.set('bugsnag.span.category', 'custom')
-    spanAttributes.set('bugsnag.span.first_class', true)
```
yousif-bugsnag, line 7: Is this right? I thought that unless it's set in SpanOptions it should be omitted?

imjoehaines: Yeah, this is deleting the line that added it to every span

yousif-bugsnag: This is what happens when you start doing code reviews before coffee

yousif-bugsnag (approved):

packages/platforms/browser/lib/span-attributes-source.ts
imjoehaines: I thought that might be the case lol

PR #224, opened 2023-06-22
Make `PageLoadPhase` spans sub-spans of the full page load span

yousif-bugsnag:
## Goal

Ensure all `PageLoadPhase` spans are direct children of the full page load span

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
@@ -1,4 +1,4 @@
-import type { SpanFactory } from '@bugsnag/core-performance'
+import type { SpanContext, SpanFactory } from '@bugsnag/core-performance'
 import { type PerformanceWithTiming } from '../on-settle/load-event-end-settler'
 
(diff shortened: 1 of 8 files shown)


packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
```diff
@@ -1,36 +1,49 @@
...
 
   if (entry) {
     if (!shouldOmitSpan(entry.unloadEventStart, entry.unloadEventEnd)) {
       spanFactory.endSpan(spanFactory.startSpan('[PageLoadPhase/Unload]' + route, {
-        startTime: entry.unloadEventStart
+        startTime: entry.unloadEventStart,
+        parentContext: pageLoadSpan,
+        makeCurrentContext: false
       }), entry.unloadEventEnd)
     }
```
gingerbenw, line 24: there's actually quite a lot of repeated code here, which could probably be abstracted into a helper function which accepts a name, start/end time and the span factory

yousif-bugsnag: good shout! updated

gingerbenw (approved):

imjoehaines (approved):

PR #225, opened 2023-06-22
[PLAT-10231] ResourceLoad spans

gingerbenw:
## Goal

Implement a new ResourceLoadPlugin in the browser client, to enable automatic instrumentation of resource load spans when there is an open span in the context stack

## Testing

Unit tests for automatic span creation with correct parentage
Unit test fetch/xhr requests do not create resource load spans
Unit test resource load spans are not created when there is no parent context
End to end tests for automated resource load spans with correct parentage

### DIFF
--- a/packages/core/lib/span-factory.ts
+++ b/packages/core/lib/span-factory.ts
@@ -51,10 +51,6 @@ export class SpanFactory {
     this.openSpans = new WeakSet<SpanInternal>()
   }
 
-  get firstSpanContext () {
-    return this.spanContextStorage.first
-  }
-
   startSpan (name: string, options: SpanOptions = {}) {
     const safeStartTime = timeToNumber(this.clock, options ? options.startTime : undefined)
     const spanId = this.idGenerator.generate(64)
--- a/packages/core/lib/span-factory.ts
+++ b/packages/core/lib/span-factory.ts
@@ -51,6 +51,10 @@ export class SpanFactory {
     this.openSpans = new WeakSet<SpanInternal>()
   }
 
+  get firstSpanContext () {
+    return this.spanContextStorage.first
+  }
+
   startSpan (name: string, options: SpanOptions = {}) {
     const safeStartTime = timeToNumber(this.clock, options ? options.startTime : undefined)
     const spanId = this.idGenerator.generate(64)
--- a/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
@@ -5,6 +5,32 @@ interface ResourceTiming extends PerformanceResourceTiming {
   responseStatus?: number // https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/responseStatus
 }
 
+export function getHTTPFlavor (protocol: string) {
+  switch (protocol) {
+    case 'http/1.0':
+      return '1.0'
+    case 'http/1.1':
+      return '1.1'
+    case 'h2':
+    case 'h2c':
+      return '2.0'
+    case 'h3':
+      return '3.0'
+    case 'spdy/1':
+    case 'spdy/2':
+    case 'spdy/3':
+      return 'SPDY'
+    default:
+      return protocol
+  }
+}
+
+function resourceLoadSupported (PerformanceObserverClass: typeof PerformanceObserver) {
+  return PerformanceObserverClass &&
+    Array.isArray(PerformanceObserverClass.supportedEntryTypes) &&
+    PerformanceObserverClass.supportedEntryTypes.includes('resource')
+}
+
 export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
   constructor (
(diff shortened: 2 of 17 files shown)


test/browser/features/resource-load-spans.feature
```diff
@@ -0,0 +1,30 @@
...
+    And I wait to receive a sampling request
+
+    When I click the element "end-span"
+    And I wait to receive 1 trace
+
+    # Custom span (parent)
+    Then the trace payload field "resourceSpans.0.scopeSpans.0.spans.2.name" equals "[Custom]/resource-load-spans"
+    And the trace payload field "resourceSpans.0.scopeSpans.0.spans.2.spanId" is stored as the value "parent_span_id"
+
+    # App bundle
```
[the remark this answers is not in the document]
gingerbenw: depending on browser versions, yes! we should get some.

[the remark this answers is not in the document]
gingerbenw: considering these spans are already covered in this feature and will be incredibly difficult to test in the page load span feature, I think this might be a bit redundant.

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+  ) {}
+
+  configure (configuration: InternalConfiguration<BrowserConfiguration>) {
+    if (!resourceLoadSupported(this.PerformanceObserverClass)) return
+
+    const observer = new this.PerformanceObserverClass((list) => {
+      const entries = list.getEntries() as ResourceTiming[]
+
+      for (const entry of entries) {
+        if (['fetch', 'xmlhttprequest'].includes(entry.initiatorType)) continue
```
imjoehaines, line 47: Little concerned about making this array & iterating through it for every entry:

```suggestion
        if (entry.initiatorType === 'fetch' || entry.initiatorType === 'xmlhttprequest') {
         continue
       }
```

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -0,0 +1,81 @@
+import { type InternalConfiguration, type Plugin, type SpanFactory } from '@bugsnag/core-performance'
+import { type BrowserConfiguration } from '../config'
+
+interface ResourceTiming extends PerformanceResourceTiming {
+  responseStatus?: number // https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/responseStatus
+}
+
+export function getHTTPFlavor (protocol: string) {
```
imjoehaines, line 8: nit: could we call this `getHttpFlavour` or `getHttpVersion`?

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+      const entries = list.getEntries() as ResourceTiming[]
+
+      for (const entry of entries) {
+        if (['fetch', 'xmlhttprequest'].includes(entry.initiatorType)) continue
+
+        const parentContext = this.spanFactory.firstSpanContext
+
+        if (parentContext) {
+          const url = new URL(entry.name)
+          const name = url.href.replace(url.search, '')
```
imjoehaines, line 53: Might be faster to clear out `url.search` rather than use `replace`:

```suggestion
          const url = new URL(entry.name)
          url.search = ''
          const name = url.href
```

packages/platforms/browser/lib/browser.ts
```diff
@@ -51,7 +51,8 @@ const BugsnagPerformance = createClient({
       performance
     ),
     new NetworkRequestPlugin(spanFactory, fetchRequestTracker, xhrRequestTracker),
-    new RouteChangePlugin(spanFactory, window.location)
+    new RouteChangePlugin(spanFactory, window.location),
+    new ResourceLoadPlugin(spanFactory, window.PerformanceObserver)
```
imjoehaines, line 55: Worth a comment that the order is important if we're relying on this being after the page load plugin for parentage

packages/core/lib/span-factory.ts
```diff
@@ -51,6 +51,10 @@ export class SpanFactory {
     this.openSpans = new WeakSet<SpanInternal>()
   }
 
+  get firstSpanContext () {
+    return this.spanContextStorage.first
+  }
```
imjoehaines, line 56: I think it'd be better to expose `spanContextStorage` to plugins in the same way the span factory is

i.e. change `options.plugins`:

```diff
-plugins: (spanFactory: SpanFactory) => Array<Plugin<C>>
+plugins: (spanFactory: SpanFactory, spanContextStorage: SpanContextStorage) => Array<Plugin<C>>
```

Alternatively make `options.spanContextStorage` not optional so `browser.ts` can pass it in directly

gingerbenw: oh, that's a much nicer idea! I can definitely see the use in it being available to other plugins - going with option 1

imjoehaines (approved):

PR #226, opened 2023-06-23
Start FullPageLoad spans earlier

yousif-bugsnag:
## Goal

Currently we start and end full page load spans after the page has settled so that we can resolve the route - however we now need to start (and end) child spans  such as resource load spans before the page has settled, so the full page load span needs to be started earlier.

This PR updates the full page load plugin to start the span as soon as the plugin is configured, and then update the name with the route once the page has settled.

The full page load span now also passes `parentContext: null` when starting the span to ensure that it is always a root span (with no parent)

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
@@ -54,7 +54,7 @@ export class FullPageLoadPlugin implements Plugin<BrowserConfiguration> {
       if (this.wasBackgrounded) return
 
       const route = configuration.routingProvider.resolveRoute(url)
-      span.name += `${route}`
+      span.name += route
 
       // Browser attributes
       span.setAttribute('bugsnag.span.category', 'full_page_load')
--- a/packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
@@ -52,7 +52,7 @@ export class FullPageLoadPlugin implements Plugin<BrowserConfiguration> {
       return
     }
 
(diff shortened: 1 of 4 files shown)


packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
```diff
@@ -47,15 +47,14 @@ export class FullPageLoadPlugin implements Plugin<BrowserConfiguration> {
...
     const url = new URL(this.location.href)
 
     this.onSettle((endTime: number) => {
       if (this.wasBackgrounded) return
 
       const route = configuration.routingProvider.resolveRoute(url)
-
-      const startTime = 0
-      const span = this.spanFactory.startSpan(`[FullPageLoad]${route}`, { startTime })
+      span.name += `${route}`
```
imjoehaines, line 57:
```suggestion
      span.name += route
```

packages/platforms/browser/tests/page-load-span-plugin.test.ts
```diff
@@ -302,6 +305,39 @@ describe('FullPageLoadPlugin', () => {
     expect(delivery).toHaveSentSpan(expect.objectContaining({ name: '[FullPageLoad]/initial-route' }))
   })
 
+  it('starts a page load span when configured and updates the route on settle', async () => {
```
imjoehaines, line 308: A better test might be to start a new span and assert it has a parent span ID, then when the full page load span ends assert its span ID matches the parent span ID

yousif-bugsnag: yeah good point. Also just realised that we're not setting `parentContext: null` when starting the page load span span, so if a span is already open the page load span will become a child (which is wrong).

Probably makes sense to add that as part of this PR

yousif-bugsnag: Updated

imjoehaines (approved):

PR #231, opened 2023-06-23
Set category and phase attributes on page load phase spans

yousif-bugsnag:
## Goal

Sets the following required attributes for page load phase spans:

`bugsnag.span.category: page_load_phase`
`bugsnag.phase: Phase`

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
@@ -12,30 +12,34 @@ export const instrumentPageLoadPhaseSpans = (
   route: string,
   parentContext: SpanContext
 ) => {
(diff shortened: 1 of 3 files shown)


packages/platforms/browser/lib/auto-instrumentation/page-load-phase-spans.ts
```diff
@@ -12,30 +12,34 @@ export const instrumentPageLoadPhaseSpans = (
   route: string,
   parentContext: SpanContext
 ) => {
-  function createPageLoadPhaseSpan (name: string, startTime: number, endTime: number) {
+  function createPageLoadPhaseSpan (phase: string, startTime: number, endTime: number) {
```
gingerbenw, line 15: wonder if it's worth making this a union? 

`type Phase = 'Unload' | 'Redirect' | 'LoadFromCache'` etc...

gingerbenw (approved):

PR #235, opened 2023-06-26
Update changelog

yousif-bugsnag:
## Goal

Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,19 @@
 # Changelog
 
-## TBD
+## v0.1.0-alpha.0 (2023-06-26)
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
 # Changelog
 
-## TBD
+## v0.1.0-alpha.0 (2023-06-26)
```
tomlongridge, line 4: Ideally we should callout the breaking change a little more – e.g. https://github.com/bugsnag/bugsnag-android-performance/blob/next/CHANGELOG.md?plain=1#L5-L6.

We don't (IMO) need a changelog for specifically the `alpha`, so if you can make sure that's added when we cut the full release that's fine.

tomlongridge (approved):

CHANGELOG.md
yousif-bugsnag: good point! I've backed out the version number change so this is now just a PR to bring the changelog up to date in advance of cutting the full release 🙂

PR #237, opened 2023-06-27
Add CDN upload script

imjoehaines:
## Goal

This is "inspired" by [the one in @bugsnag/js](https://github.com/bugsnag/bugsnag-js/blob/21836527fd91b01020135bbb94ccf93ecdaa1c59/bin/cdn-upload) but updated to use AWS' v3 packages

Run it in much the same way:

```sh
# with AWS credentials in environment variables
$ npm run cdn-upload
```

Tested against the sandbox AWS bucket/cloudfront distribution

### DIFF
--- a/bin/cdn-upload
+++ b/bin/cdn-upload
@@ -5,12 +5,20 @@ const semver = require('semver')
 const { S3Client, PutObjectCommand } = require('@aws-sdk/client-s3')
 const { CloudFrontClient, CreateInvalidationCommand } = require('@aws-sdk/client-cloudfront')
 
+function getEnvironmentVariable (name) {
+  if (process.env[name]) {
+    return process.env[name]
+  }
+
+  throw new Error(`Missing environment variable '${name}'`)
+}
+
 const packageJson = require(`${__dirname}/../packages/platforms/browser/package.json`)
 const packageName = packageJson.name
 const packageVersion = packageJson.version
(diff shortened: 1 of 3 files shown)


yousif-bugsnag (approved):

bin/cdn-upload
```diff
@@ -0,0 +1,92 @@
...
+const semver = require('semver')
+const { S3Client, PutObjectCommand } = require('@aws-sdk/client-s3')
+const { CloudFrontClient, CreateInvalidationCommand } = require('@aws-sdk/client-cloudfront')
+
+const packageJson = require(`${__dirname}/../packages/platforms/browser/package.json`)
+const packageName = packageJson.name
+const packageVersion = packageJson.version
+const bucketName = process.env.BUCKET_NAME || 'bugsnagcdn'
+const distributionId = process.env.DISTRIBUTION_ID || 'E205JDPNKONLN7'
+const region = process.env.AWS_REGION || 'us-east-1'
```
gingerbenw, line 13: I'm sure we've already discussed these before - but would it be better to only use the environment variables?

imjoehaines: Maybe; I think this is safe because you can't have AWS credentials laying around in your environment anymore, so it's always a conscious effort to run this script with the ability to upload to the real CDN

That said, the release script could just set them automatically so I don't think there's a downside

gingerbenw (approved):

PR #240, opened 2023-06-28
Page load and Route change span attributes

gingerbenw:
## Goal

To set url and title attributes when a page load or route change span settles, rather than when it's created

## Testing

Updated unit tests


gingerbenw:
Resolves issue mention in PLAT-10342

packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
```diff
@@ -20,14 +19,25 @@ export class RouteChangePlugin implements Plugin<BrowserConfiguration> {
         startTime: options ? options.startTime : undefined
       })
 
+      const url = getAbsoluteUrl(route, this.document.baseURI)
```
imjoehaines, line 22: @gingerbenw `route` is an arbitrary string, not necessarily a URL or path

e.g. it's valid to have a route like `user #1234` instead of `/users/1234`

gingerbenw: good point! hence providing the route resolution function! I shall raise a new PR

PR #245, opened 2023-06-30
Handle invalid inputs to startSpan

yousif-bugsnag:
## Goal
This PR adds validation to inputs when starting spans, both via `BugsnagPerformance.startSpan` and the `startRouteChangeSpan` callback in routing providers.

Takes a similar approach to config validation, adding a generic validation function and a core span options schema that can be extended by different platforms/plugins (see `route-change-plugin.ts`)

If an input is invalid, a warning is logged and the default value is used instead. For all SpanOptions this is `undefined` and for span names the invalid input is converted to a string. 

## Testing
Updated and added unit tests for input validation and reorganised the span and span factory unit tests to split out tests for invalid options from the actual implementation of each option

### DIFF
--- a/packages/core/lib/span-factory.ts
+++ b/packages/core/lib/span-factory.ts
@@ -5,7 +5,7 @@ import { type Logger } from './config'
 import { type IdGenerator } from './id-generator'
 import { type Processor } from './processor'
 import { type ReadonlySampler } from './sampler'
-import { type Span, SpanInternal, type SpanOptions, validateSpanOptions } from './span'
+import { SpanInternal, type Span, type SpanOptions } from './span'
 import { type SpanContextStorage } from './span-context'
 import { timeToNumber } from './time'
 import { isSpanContext } from './validation'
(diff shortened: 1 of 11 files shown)


packages/core/lib/span-factory.ts
```diff
@@ -51,33 +51,35 @@ export class SpanFactory {
...
 
-    const span = new SpanInternal(spanId, traceId, name, safeStartTime, attributes, parentSpanId)
+    const span = new SpanInternal(spanId, traceId, cleanOptions.name, safeStartTime, attributes, parentSpanId)
 
     // don't track spans that are started while the app is backgrounded
     if (this.isInForeground) {
       this.openSpans.add(span)
 
-      if (!options || options.makeCurrentContext !== false) {
+      if (!cleanOptions || cleanOptions.makeCurrentContext !== false) {
```
gingerbenw, line 82: with the new validation function, clean options will never be falsey

packages/core/tests/span-factory.test.ts
```diff
@@ -1,120 +1,241 @@
...
-      const span = spanFactory.startSpan('name', { isFirstClass: true })
-
-      // @ts-expect-error 'attributes' is private but very awkward to test otherwise
-      expect(span.attributes.attributes.get('bugsnag.span.first_class')).toBe(true)
+        // new root span should have a new trace ID and no parentSpanId
+        expect(delivery).toHaveSentSpan(expect.objectContaining({
+          name: 'new root span',
+          parentSpanId: rootSpan.id,
+          traceId: rootSpan.traceId
+        }))
```
gingerbenw, line 104: comment does not match the assertion

gingerbenw (approved):
few comments, but otherwise looks good!

packages/core/tests/span-factory.test.ts
yousif-bugsnag: whoops!

PR #247, opened 2023-07-03
Validate bugsnag.browser.page.url

gingerbenw:
## Goal

Ensure `bugsnag.browser.page.url` attribute accurately reflects the new url after a route change

## Changeset

Added a new url parameter to `startRouteChangeSpan` callback in `listenForRouteChanges`

## Testing

Updated end to end tests

### DIFF
--- a/packages/platforms/browser/lib/routing-provider.ts
+++ b/packages/platforms/browser/lib/routing-provider.ts
@@ -1,7 +1,7 @@
 import { isObject, type Span, type SpanOptions } from '@bugsnag/core-performance'
 
 export type RouteChangeSpanOptions = Omit<SpanOptions, 'isFirstClass'>
-export type StartRouteChangeCallback = (url: URL, trigger: string, options?: RouteChangeSpanOptions) => Span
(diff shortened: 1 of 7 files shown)


packages/platforms/browser/lib/routing-provider.ts
```diff
@@ -1,7 +1,7 @@
 import { isObject, type Span, type SpanOptions } from '@bugsnag/core-performance'
 
 export type RouteChangeSpanOptions = Omit<SpanOptions, 'isFirstClass'>
-export type StartRouteChangeCallback = (route: string, trigger: string, options?: RouteChangeSpanOptions) => Span
+export type StartRouteChangeCallback = (url: string, route: string, trigger: string, options?: RouteChangeSpanOptions) => Span
```
imjoehaines, line 4: If this accepts a `url` then it doesn't need `route` since we can call `resolveRoute` to get that

It also might be worth making it `URL | string`?

imjoehaines (approved):

gingerbenw:
needed some extra work to handle not passing a valid URL, and updated to accept a string.

PR #251, opened 2023-07-05
Run desktop browser tests against the CDN build in full pipeline

imjoehaines:
## Goal

This PR extends the "full build" pipeline to run the basic pipeline against the CDN build

The full pipeline uploads the basic pipeline with the `USE_CDN_BUILD` environment variable defined, which ultimately causes the fixtures to use the CDN build instead of `npm pack`-ing the packages and installing those

The CDN bundle is then imported in the existing test fixture JS files so that it reflects real-world usage (as opposed to bundling the CDN build into the fixture bundles)

This required a lot of supporting changes to the way we prepare and build the fixtures but is all controlled by the `USE_CDN_BUILD` environment variable in the existing `build_packages.rb` script & rollup config

## Testing

A manually triggered full build passes: https://buildkite.com/bugsnag/bugsnag-js-performance/builds/1662

You can tell if the CDN build is being used because it logs a message at `INFO` level:

image  image

The resource load test has been split into two (one for the NPM build and one for the CDN build because the CDN bundle is an extra resource load) so you can also look at which version has been run

### DIFF
--- a/test/browser/features/support/build-packages.rb
+++ b/test/browser/features/support/build-packages.rb
@@ -1,15 +1,25 @@
-# set the SKIP_BUILD_PACKAGES environment variable to disable building
-return if ENV.key?("SKIP_BUILD_PACKAGES")
+def environment_variable_enabled?(name)
+  value = ENV[name]
 
-require 'open3'
-require 'logger'
+  value == "1" || value == "true"
+end
+
+# set the SKIP_BUILD_PACKAGES environment variable to disable building
+return if environment_variable_enabled?("SKIP_BUILD_PACKAGES")
+
+require "json"
+require "open3"
+require "logger"
 
 # this file is run by Maze Runner automatically (where $logger is defined) and
 # by the browser dockerfile when building for CI (where $logger is NOT defined)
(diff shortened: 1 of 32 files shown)


test/browser/features/support/build-packages.rb
```diff
@@ -55,24 +65,55 @@ def run(command)
...
+      # in CDN mode copy the bundles & sourcemaps (for debugging) into the
+      # fixture directory
+      run("cp build/bugsnag-performance*.js* #{FIXTURES_DIRECTORY}/packages/")
     end
   end
 
   Dir.chdir(FIXTURES_DIRECTORY) do
-    run("npm install --no-package-lock --no-save *.tgz")
-    run("npm run build --workspaces")
+    # backup the package json so we can undo changes later to stop git from
```
gingerbenw, line 81: to stop git from what!? :D

gingerbenw (approved):

test/browser/features/support/build-packages.rb
imjoehaines: To stop it _from_ 🙄

PR #255, opened 2023-07-06
[PLAT-10560] Remove empty http.flavor span attribute

gingerbenw:
## Goal

To only add an `http.flavor` span attribute if it is a usable value.

## Testing

Updated unit tests

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
@@ -7,6 +7,8 @@ interface ResourceTiming extends PerformanceResourceTiming {
 
 export function getHttpVersion (protocol: string) {
   switch (protocol) {
+    case '':
+      return undefined
     case 'http/1.0':
       return '1.0'
     case 'http/1.1':
@@ -21,7 +23,7 @@ export function getHttpVersion (protocol: string) {
     case 'spdy/3':
       return 'SPDY'
     default:
-      return undefined
+      return protocol
   }
 }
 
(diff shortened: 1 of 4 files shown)


yousif-bugsnag (approved):

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -7,6 +7,8 @@ interface ResourceTiming extends PerformanceResourceTiming {
 
 export function getHttpVersion (protocol: string) {
   switch (protocol) {
```
imjoehaines, line 9: Would changing the `default` instead make sense?

gingerbenw: probably! I guess it depends if we want to exclude all of the other [possibilities of nextHopProtocol](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids) ?

@gareththackeray any thoughts on this?

gingerbenw: The product design lists the values for `http.flavor` as `1.0 | 1.1 | 2.0 | 3.0 | SPDY | QUIC` so have gone ahead and updated it to return `undefined` for anything outside of this range.

gareththackeray: I think use watever you get back if it has a value

gingerbenw: so, just to clarify - we want to map the values to those as described in the PD (such as `h2` and h2c` to `2.0`) and anything outside of that range, we use as it is? like `xmpp` or `stun.turn`

gareththackeray: Yes I think so

gareththackeray (approved):

PR #257, opened 2023-07-07
Correctly report sampling header

imjoehaines:
## Goal

Correctly report the `Bugsnag-Span-Sampling` header

This required a lot of changes as the current delivery system didn't have enough information to calculate the header

TO fix this, I've introduced a new `TracePayload` type that contains both a `body` and `headers`:

```ts
interface TracePayload {
  body: DeliveryPayload
  headers: {
    'Bugsnag-Api-Key': string
    'Content-Type': 'application/json'
    'Bugsnag-Span-Sampling': string
    // we don't add 'Bugsnag-Sent-At' in the TracePayloadEncoder so that retried
    // payloads get a new value each time delivery is attempted
    // therefore it's 'undefined' when passed to delivery, which adds a value
    // immediately before initiating the request
    'Bugsnag-Sent-At'?: string
  }
}
```

This is created by a new `TracePayloadEncoder`, which takes an array of `SpanEnded` and returns a `TracePayload`

The `BatchProcessor` then uses the `TracePayloadEncoder` to construct the payload before passing it to the delivery system

I originally intended for delivery to do the payload encoding, but this causes a lot of issues (mainly in our tests):

- many of our tests rely on asserting against delivery payloads
- the `TracePayloadEncoder` requires a configuration, clock & resource attribute source to construct the payload. Adding this to every place we use `InMemoryDelivery` is very awkward
- it's arguably better for encoding to be done in core before being passed to delivery; otherwise every platform's delivery needs to repeat the encoding step

### DIFF
--- a/packages/core/lib/delivery.ts
+++ b/packages/core/lib/delivery.ts
@@ -1,8 +1,10 @@
-import { type JsonAttribute } from './attributes'
+import { type Configuration, type InternalConfiguration } from './config'
+import { type Clock } from './clock'
+import { type JsonAttribute, type ResourceAttributeSource } from './attributes'
 import { type JsonEvent } from './events'
-import { type Kind } from './span'
+import { type Kind, type SpanEnded, spanToJson } from './span'
 
-export type DeliveryFactory = (apiKey: string, endpoint: string) => Delivery
+export type DeliveryFactory = (endpoint: string) => Delivery
 
 export type ResponseState = 'success' | 'failure-discard' | 'failure-retryable'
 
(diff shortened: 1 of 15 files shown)


packages/core/lib/delivery.ts
```diff
@@ -44,6 +46,88 @@ export interface DeliverySpan {
...
+    const pairs = Array(rawProbabilities.length)
+
+    for (let i = 0; i < rawProbabilities.length; ++i) {
+      const rawProbability = rawProbabilities[i]
+
+      pairs[i] = `${rawProbability}:${spanCounts[rawProbability]}`
+    }
+
+    return pairs.join(';')
+  }
```
gingerbenw, line 128: 👍🏻

gingerbenw (approved):
looking good!

gingerbenw:
@yousif-bugsnag  think the code coverage bot needs a little work - these should probably be green pips?

PR #258, opened 2023-07-10
Add `bugsnag.sampling.p` attribute

imjoehaines:
## Goal

Add the `bugsnag.sampling.p` attribute which stores a span's probability value

This is set automatically when a `SpanInternal` is ended and when its probability is updated


packages/core/lib/attributes.ts
```diff
@@ -60,9 +60,12 @@ export function attributeToJson (key: string, attribute: SpanAttribute): JsonAtt
       if (Number.isNaN(attribute) || !Number.isFinite(attribute)) {
         return undefined
       }
-      if (Number.isInteger(attribute)) {
+
+      // 'bugsnag.sampling.p' must always be sent as a doubleValue
+      if (key !== 'bugsnag.sampling.p' && Number.isInteger(attribute)) {
```
imjoehaines, line 65: This is a bit of a hack but we don't have a way to tell the attribute encoding that something needs to be of a specific type

Maybe this kind of auto-encoding was a bad idea?

gingerbenw: It definitely made sense when we initially implemented it!

Not too concerned about it - but worth refactoring if we end up adding _more_ edge cases, I'd say?

gingerbenw (approved):

yousif-bugsnag (approved):

PR #261, opened 2023-07-12
[PLAT-10511] Remove networkInstrumentationIgnoreUrls option

gingerbenw:
## Goal

To remove an option that will be covered by new functionality in the network span controls callback.

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
@@ -8,7 +8,7 @@ import { type SpanFactory, type Plugin, type InternalConfiguration } from '@bugs
 import { type BrowserConfiguration } from '../config'
 
 export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
-  private ignoredUrls: RegExp[] = []
+  private configEndpoint: string = ''
 
   constructor (
     private spanFactory: SpanFactory,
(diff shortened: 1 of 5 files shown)


packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -18,10 +18,7 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
 
   configure (configuration: InternalConfiguration<BrowserConfiguration>) {
     if (configuration.autoInstrumentNetworkRequests) {
-      this.ignoredUrls = configuration.networkInstrumentationIgnoreUrls.map(
-        (url: string | RegExp): RegExp => typeof url === 'string' ? RegExp(url) : url
-      ).concat(RegExp(configuration.endpoint))
-
+      this.ignoredUrls = [RegExp(configuration.endpoint)]
```
imjoehaines, line 21: This doesn't need to be an array (or regex really) anymore

yousif-bugsnag (approved):

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -48,6 +45,6 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
   }
 
   private shouldTrackRequest (startContext: RequestStartContext): boolean {
-    return !this.ignoredUrls.some(url => url.test(startContext.url))
+    return !startContext.url.includes(this.configEndpoint)
```
imjoehaines, line 48: Does this need to be an `includes`? We make requests to the exact endpoint so it feels like `===` would be fine?

gingerbenw: that's a good point - I think I was assuming the possibility of additional paths or query strings. strict equality would be fine here.

PR #263, opened 2023-07-13
[PLAT-10604] Implement networkRequestCallback in Network Span Plugin

gingerbenw:
## Goal

use the `networkRequestCallback` plugin in the `network-span-plugin` in order to use the modified url or prevent creation of spans.

## Testing

Unit tests to ensure modified URLs are used by `network-span-plugin`
Unit tests to ensure spans are not created when `networkRequestCallback` returns `null`

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
@@ -1,4 +1,4 @@
-import { type InternalConfiguration, type Plugin, type SpanFactory } from '@bugsnag/core-performance'
+import { type InternalConfiguration, type Logger, type Plugin, type SpanFactory } from '@bugsnag/core-performance'
 import { type BrowserConfiguration } from '../config'
 import { defaultNetworkRequestCallback, type NetworkRequestCallback } from '../network-request-callback'
 import {
@@ -11,6 +11,7 @@ import {
 export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
   private configEndpoint: string = ''
   private networkRequestCallback: NetworkRequestCallback = defaultNetworkRequestCallback
+  private logger: Logger = { debug: console.debug, warn: console.warn, info: console.info, error: console.error }
 
   constructor (
     private spanFactory: SpanFactory,
@@ -19,6 +20,8 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
   ) {}
 
   configure (configuration: InternalConfiguration<BrowserConfiguration>) {
+    this.logger = configuration.logger
+
     if (configuration.autoInstrumentNetworkRequests) {
       this.configEndpoint = configuration.endpoint
       this.xhrTracker.onStart(this.trackRequest)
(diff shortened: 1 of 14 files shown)


packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -24,19 +26,24 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
+    if (!networkRequestInfo) return
+
     const span = this.spanFactory.startSpan(
       `[HTTP]/${startContext.method.toUpperCase()}`,
       { startTime: startContext.startTime, makeCurrentContext: false }
     )
 
     span.setAttribute('bugsnag.span.category', 'network')
-    span.setAttribute('http.url', startContext.url)
+    span.setAttribute('http.url', networkRequestInfo.url)
```
imjoehaines, line 46: What if someone returns a truthy value that doesn't have a `url` property? e.g. `true` or a random object

packages/platforms/browser/lib/network-request-callback.ts
```diff
@@ -1,6 +1,6 @@
 export interface NetworkRequestInfo {
   url: string
-  readonly type: string
+  readonly type: 'fetch' | 'xmlhttprequest'
```
yousif-bugsnag, line 3: if this is going to be used by the resource load plugin then it probably needs to be a more permissive type that corresponds to `PerformanceResourceTiming.initiatorType`  - it's only `RequestStartContext.type` that's restricted to `fetch`/`xmlhttprequest`

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -17,23 +20,35 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
+
     const span = this.spanFactory.startSpan(
       `[HTTP]/${startContext.method.toUpperCase()}`,
       { startTime: startContext.startTime, makeCurrentContext: false }
     )
 
+    if (typeof networkRequestInfo.url === 'string') {
+      span.setAttribute('http.url', networkRequestInfo.url)
+    } else {
+      this.logger.warn(`expected url to be a string, got ${typeof networkRequestInfo.url}, http.url attribute discarded.`)
```
gingerbenw, line 48: not sure about the wording here @tomlongridge

tomlongridge: Maybe some context would be useful - "expected url to be a string following network request callback"

yousif-bugsnag (approved):

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -17,23 +20,35 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
+
+    if (!networkRequestInfo) return
+
     const span = this.spanFactory.startSpan(
       `[HTTP]/${startContext.method.toUpperCase()}`,
       { startTime: startContext.startTime, makeCurrentContext: false }
     )
 
+    if (typeof networkRequestInfo.url === 'string') {
+      span.setAttribute('http.url', networkRequestInfo.url)
```
imjoehaines, line 46: I don't think this makes sense; is a network span without a URL ever going to be useful?

imjoehaines: The server also expects this to always be set for both network and resource spans, so we can't make it optional

gingerbenw: updated! if we return anything except a string we log a warning and don't create the span

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -17,24 +20,36 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
   }
 
   private trackRequest = (startContext: RequestStartContext): RequestEndCallback | undefined => {
     if (!this.shouldTrackRequest(startContext)) return
 
+    const networkRequestInfo = this.networkRequestCallback({ url: startContext.url, type: startContext.type })
+
+    if (networkRequestInfo === null) return
+
+    if (typeof networkRequestInfo.url !== 'string') {
```
imjoehaines, line 40: This will crash if `networkRequestInfo` is `undefined`

imjoehaines (approved):

PR #268, opened 2023-07-17
Calculate time origin on create as well as visibility change

yousif-bugsnag:
## Goal

It's possible the performance clock may be shared between different browser tabs running in the same process, and could therefore already have diverged from the wall clock when the client is created.

This PR attempts to mitigate this by updating the browser clock implementation to check for wall clock divergence and calculate a time origin when the clock is created as well as when the app returns to the foreground.


gingerbenw (approved):
looks good, but needs a changelog entry :)

PR #272, opened 2023-07-19
Refactor span attributes source

gingerbenw:
## Goal

Refactor span attributes source to enable adding default platform attributes depending on configuration 

## Design

Request permitted attributes from span attribute source when a span is ended to ensure latest value is used 
Configure span attribute source to permit attributes to be added

## Testing

Unit test configure and request methods.


packages/core/lib/span-factory.ts
```diff
@@ -65,7 +65,7 @@ export class SpanFactory {
     const parentSpanId = parentContext ? parentContext.id : undefined
     const traceId = parentContext ? parentContext.traceId : this.idGenerator.generate(128)
 
-    const attributes = new SpanAttributes(this.spanAttributesSource())
+    const attributes = new SpanAttributes(new Map())
```
yousif-bugsnag, line 68: could the span attributes source just return a `SpanAttributes` instance instead?

```ts
const attributes = this.spanAttributesSource.requestAttributes()
```

or even pass the configuration down through like we do for resource attributes: https://github.com/bugsnag/bugsnag-js-performance/blob/8460c1f2c1491db12937ccf4b5d12d078473ed23/packages/core/lib/delivery.ts#L79

gingerbenw: unfortunately we need to get the latest `title` and `url` attributes when the span is _ended_ as there may be all sorts of routing and dom updating things going on during the process, so these have to be added on end.

yousif-bugsnag: oh yes good point

yousif-bugsnag (approved):

PR #274, opened 2023-07-21
[PLAT-10603] Implement networkRequestCallback in resource loads

gingerbenw:
## Goal

Modify URLs or prevent delivery of spans in resource load spans

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
@@ -53,8 +53,10 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
 
         const parentContext = this.spanContextStorage.first
 
-        if (parentContext) {
-          const url = new URL(entry.name)
+        const networkRequestInfo = configuration.networkRequestCallback({ url: entry.name, type: entry.initiatorType })
+
+        if (parentContext && networkRequestInfo) {
+          const url = new URL(networkRequestInfo.url)
           url.search = ''
           const name = url.href
 
@@ -65,7 +67,7 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
           })
 
           span.setAttribute('bugsnag.span.category', 'resource_load')
-          span.setAttribute('http.url', entry.name)
+          span.setAttribute('http.url', networkRequestInfo.url)
 
           const httpFlavor = getHttpVersion(entry.nextHopProtocol)
           if (httpFlavor) {
--- a/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
@@ -56,7 +56,11 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
         if (parentContext) {
           const networkRequestInfo = configuration.networkRequestCallback({ url: entry.name, type: entry.initiatorType })
 
-          if (!networkRequestInfo || typeof networkRequestInfo.url !== 'string') return
+          if (!networkRequestInfo) return
+
+          if (typeof networkRequestInfo.url !== 'string') {
+            configuration.logger.warn(`expected url to be a string following network request callback, got ${typeof networkRequestInfo.url}`); return
+          }
 
           const url = new URL(networkRequestInfo.url)
           url.search = ''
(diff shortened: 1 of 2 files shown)


packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -53,8 +53,10 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
 
         const parentContext = this.spanContextStorage.first
 
-        if (parentContext) {
-          const url = new URL(entry.name)
+        const networkRequestInfo = configuration.networkRequestCallback({ url: entry.name, type: entry.initiatorType })
```
imjoehaines, line 56: Does it make sense to call this if there's no `parentContext` and we won't be making a span?

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -53,8 +53,10 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
...
+
+        if (parentContext && networkRequestInfo) {
+          const url = new URL(networkRequestInfo.url)
```
imjoehaines, line 59: What if `networkRequestInfo.url` isn't a string?

gingerbenw: I suppose we probably need to log a warning and not create the span... we could add a try/catch to the URL constructor which would also gracefully handle an incomplete URL?

gingerbenw: actually, that's a little over-complex and if we're monkey patching a native request, the user will likely already have thrown an error in their application from trying to make a request with an invalid endpoint. updated!

imjoehaines: The logic should be the same as in #263

A try/catch for invalid URLs makes sense too, though the URL constructor expects a full URL so it can't just be a relative path. Should the name still use `entry.name`?

gingerbenw: I don't we should directly use `entry.name` as that's the property that the user may have removed personal information from

gingerbenw: i.e. path parameters with a user id, or an api key.

imjoehaines: OK, makes sense; we'll need to handle `networkRequestInfo.url` being invalid here then

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -54,9 +54,22 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
...
+            configuration.logger.warn(`expected url to be a string following network request callback, got ${typeof networkRequestInfo.url}`); return
+          }
+
+          let name = ''
+          try {
+            const url = new URL(networkRequestInfo.url)
+            url.search = ''
+            name = url.href
+          } catch (err) {
+            configuration.logger.warn('unable to parse url'); return
```
gingerbenw, line 71: @gareththackeray would you know what type of wording to use in this scenario? where a user has returned a string from the `networkRequestCallback` but we are unable to construct a URL with it.

gareththackeray: How about:
> Unable to parse URL returned from networkRequestCallback.

And include the URL in the message?

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -54,9 +54,22 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
...
         if (parentContext) {
-          const url = new URL(entry.name)
-          url.search = ''
-          const name = url.href
+          const networkRequestInfo = configuration.networkRequestCallback({ url: entry.name, type: entry.initiatorType })
+
+          if (!networkRequestInfo) return
+
+          if (typeof networkRequestInfo.url !== 'string') {
+            configuration.logger.warn(`expected url to be a string following network request callback, got ${typeof networkRequestInfo.url}`); return
```
imjoehaines, line 62: Can you put these returns on their own lines so they're harder to miss?

imjoehaines (approved):

packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
gingerbenw: I did put it there actually, but the prettier plugin seemed to have other ideas - I'll try again.

PR #275, opened 2023-07-24
[PLAT-10617] Only track network requests over http(s)

gingerbenw:
## Goal

Prevent errors in the pipeline from sending spans caused by network requests over additional protocols to http and https.

## Design

Prevent spans being created with urls starting with anything other than http and https, or using a relative path.

## Testing

Added unit tests for positive and negative scenarios

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
@@ -8,6 +8,8 @@ import {
   type RequestTracker
 } from '../request-tracker/request-tracker'
 
+const permittedPrefixes = ['http://', 'https://', '/', './', '../']
+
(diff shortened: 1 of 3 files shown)


packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -60,6 +60,7 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
   }
 
   private shouldTrackRequest (startContext: RequestStartContext): boolean {
-    return startContext.url !== this.configEndpoint
+    const permittedPrefixes = ['http://', 'https://', '/', './', '../']
```
imjoehaines, line 63: This makes a new array every time `shouldTrackRequest` is called; can we define this in the top-level so we only make the array once?

imjoehaines (approved):

PR #276, opened 2023-07-24
[PLAT-10639] Use pathname when resolved route is empty

gingerbenw:
## Goal

To prevent empty attributes in route change spans

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
@@ -2,6 +2,7 @@ import { coreSpanOptionSchema, isString, validateSpanOptions, type InternalConfi
 import { type BrowserConfiguration } from '../config'
 import { type RouteChangeSpanOptions } from '../routing-provider'
 import { getPermittedAttributes } from '../send-page-attributes'
+import { defaultRouteResolver } from '../default-routing-provider'
 
 // exclude isFirstClass from the route change option schema
 const { startTime, parentContext, makeCurrentContext } = coreSpanOptionSchema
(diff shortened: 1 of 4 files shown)


packages/platforms/browser/lib/default-routing-provider.ts
```diff
@@ -2,7 +2,7 @@ import { type OnSettle } from './on-settle'
 import getAbsoluteUrl from './request-tracker/url-helpers'
 import { type RouteResolver, type RoutingProvider, type StartRouteChangeCallback } from './routing-provider'
 
-const defaultRouteResolver: RouteResolver = (url: URL) => url.pathname
+const defaultRouteResolver: RouteResolver = (url: URL) => [null, undefined, ''].includes(url.pathname) ? url.toString() : url.pathname
```
imjoehaines, line 5: This is making a new array and iterating through it each time it's called

This would be more efficient and it's less code too:

```ts
(url: URL) => url.pathname || url.toString()
```

imjoehaines (commented):
Can you add some tests for this?

I'm not sure this makes sense conceptually though; isn't an empty path equivalent to `/`?

gingerbenw:
> I'm not sure this makes sense conceptually though; isn't an empty path equivalent to `/`?

That's a good point... @gareththackeray should the route be `/`?

packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
```diff
@@ -31,7 +31,9 @@ export class RouteChangePlugin implements Plugin<BrowserConfiguration> {
   configure (configuration: InternalConfiguration<BrowserConfiguration>) {
     if (!configuration.autoInstrumentRouteChanges) return
 
-    let previousRoute = configuration.routingProvider.resolveRoute(new URL(this.location.href))
+    const previousUrl = new URL(this.location.href)
+    let previousRoute = configuration.routingProvider.resolveRoute(previousUrl) || previousUrl.pathname
```
gareththackeray, line 35: Could / should we have a function to do `configuration.routingProvider.resolveRoute(url) || url.pathname` since it's used twice (i.e. also on line 73)?  (An implementation point rather than producty so feel free to ignore!)

gareththackeray (approved):
Functionally LGTM

packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
imjoehaines: It should probably just fallback to the `defaultRouteResolver` as otherwise we'd have the same bug if `previousUrl.pathname` was empty

PR #280, opened 2023-07-28
[PLAT-10646] Oversized payload handling

gingerbenw:
## Goal

Do not retry oversized payloads when connection terminates, add content length header for future usage by the pipeline

## Changeset

- Remove all payloads that fail due to a terminated connection if they are oversized
- Add Bugsnag-Uncompressed-Content-Length to communicate size to server for improved handling in future
- Use the uncompressed size when calculating the 1mb limit

## Testing

Unit test to check an oversized payload is not retried when there is a connection error

### DIFF
--- a/packages/platforms/browser/lib/delivery.ts
+++ b/packages/platforms/browser/lib/delivery.ts
@@ -46,7 +46,6 @@ function createBrowserDeliveryFactory (
         const body = JSON.stringify(payload.body)
 
         payload.headers['Bugsnag-Sent-At'] = clock.date().toISOString()
-        payload.headers['Bugsnag-Uncompressed-Content-Length'] = body.length.toString()
 
         try {
           const response = await fetch(endpoint, {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Added
-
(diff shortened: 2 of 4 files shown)


packages/platforms/browser/lib/delivery.ts
```diff
@@ -43,13 +43,16 @@ function createBrowserDeliveryFactory (
   return function browserDeliveryFactory (endpoint: string): Delivery {
     return {
       async send (payload: TracePayload) {
+        const body = JSON.stringify(payload.body)
+
         payload.headers['Bugsnag-Sent-At'] = clock.date().toISOString()
+        payload.headers['Bugsnag-Uncompressed-Content-Length'] = body.length.toString()
```
imjoehaines, line 49: This doesn't seem necessary because we don't compress the payload so would just be a repeat of `Content-Length`

It's not an accurate size though because:

- browsers can and will mutate the body before starting the request, e.g. Safari applies unicode normalisation so the body here isn't necessarily the same as the body that gets sent in the request
- `fetch` encodes strings as UTF-8 but JS strings are UTF-16
- `String#length` returns the number of UTF-16 code units in the string, not the number of bytes

I think it would be better not to add this header as the backend can use `Content-Length` for the same purpose and that is calculated by the browser based on the actual request body

imjoehaines (approved):

PR #281, opened 2023-08-01
Create delivery-fetch package

yousif-bugsnag:
## Goal

Extracts the current delivery mechanism (and associated tests) from the browser package into a separate package called `@bugsnag/delivery-fetch-performance`


packages/delivery-fetch/lib/delivery.ts
```diff
@@ -25,13 +25,13 @@ function samplingProbabilityFromHeaders (headers: Headers): number | undefined {
...
 
-function createBrowserDeliveryFactory (
+function createFetchDeliveryFactory (
   fetch: Fetch,
   backgroundingListener: BackgroundingListener,
   clock: Clock
 ): DeliveryFactory {
   // we set fetch's 'keepalive' flag if the app is backgrounded/terminated so
-  // that we can flush the last batch - without 'keepalive' the browser can
+  // that we can flush the last batch - without 'keepalive' a browser may
```
gingerbenw, line 34: browser, client? platform?

gingerbenw (approved):

packages/delivery-fetch/lib/delivery.ts
yousif-bugsnag: yeah, wasn't sure about this part as really this is browser-specific functionality that needs to remain in the shared package. Is it worth making `backgroundingListener` optional here as well as updating the comment?

gingerbenw: optional seems cleaner to me, rather than forcing in a backgrounding listener that doesn't do anything on particular platforms.

PR #283, opened 2023-08-02
Improve end to end test reliability

gingerbenw:
## Goal

Improve reliability of test fixture due to failures [like this](https://buildkite.com/bugsnag/bugsnag-js-performance/builds/1860#0189b438-f28d-4f79-a7db-696facd729a8) where the batch was too small to include the span we were asserting against


yousif-bugsnag (approved):
LGTM

PR #285, opened 2023-08-07
feat: add ReactRouterRoutingProvider

djskinner:
## Goal

Add a routing provider for react-router v6 to make it easy get page load spans that match the route definitions in a react-router application.

## Design

```
const basename = '/app'

const routes = [
    {
        path: '/',
        element: <Root />,
        children: [
          {
              path: 'contacts/:contactId',
              element: <Contact />,
          },
        ]
    },
]

const router = createBrowserRouter(routes, { basename })

BugsnagPerformance.start({
    apiKey,
    endpoint,
    maximumBatchSize: 13,
    batchInactivityTimeoutMs: 5000,
    autoInstrumentNetworkRequests: false,
    routingProvider: new ReactRouterRoutingProvider(routes, basename)
})


ReactDOM.createRoot(document.getElementById('root')).render(
  <React.StrictMode>
    <RouterProvider router={router} />
  </React.StrictMode>,
)
```

This approach provides a simple API that make it easy to integration with a react-router application when configured in the recommended way.

## Changeset

- add ReactRouterRoutingProvider

## Testing

- Manual tests
- Unit coverage
- e2e coverage


test/browser/features/fixtures/packages/rollup.config.mjs
```diff
@@ -24,8 +26,15 @@ export default {
     ...(isCdnBuild ? cdnOutputOptions : {}),
   },
   plugins: [
-    nodeResolve({ browser: true, jail: path.resolve(`${__dirname}/..`) }),
-    commonjs()
+    nodeResolve({ browser: true, jail: path.resolve(`${__dirname}/..`), extensions: ['.mjs', '.js', '.json', '.node', '.jsx'] }),
+    commonjs(),
+    babel({ babelHelpers: 'bundled' }),
```
djskinner, line 31: required to parse the jsx

test/browser/features/fixtures/packages/rollup.config.mjs
```diff
@@ -24,8 +26,15 @@ export default {
...
+    replace({
+      preventAssignment: true,
+      values: {
+        'process.env.NODE_ENV': JSON.stringify('production'),
+      },
+    })
```
djskinner, line 37: required when using react

packages/platforms/browser/lib/routing-provider/react-router-routing-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+  ]
+}
+
+export const createReactRouterRoutingProvider = (onSettle: OnSettle, location: Location) => {
+  const DefaultRoutingProvider = createDefaultRoutingProvider(onSettle, location)
+
+  return class ReactRouterRoutingProvider extends DefaultRoutingProvider {
+    constructor (routes: RouteObject[], basename?: string) {
+      function resolveRoute (url: URL): string {
+        return flattenRoutes(routes).find((fullRoutePath) => url.pathname.replace(basename ?? '', '').match(stringToRegexp(fullRoutePath))) || '/'
```
djskinner, line 27: I'm not sure whether ` || '/'` is best here. Is there a better thing to do if we were unable to resolve the route? Return `url` perhaps?

packages/platforms/browser/lib/index.ts
```diff
@@ -1,4 +1,4 @@
-export { DefaultRoutingProvider, default, onSettle } from './browser'
+export { DefaultRoutingProvider, default, onSettle, ReactRouterRoutingProvider } from './browser'
 export { type BrowserConfiguration } from './config'
-export { type RouteResolver, type RoutingProvider } from './routing-provider'
+export { type RouteResolver, type RoutingProvider, type StartRouteChangeCallback } from './routing-provider'
```
djskinner, line 3: this type is referred to in the docs but wasn't being exported https://docs.bugsnag.com/performance/integration-guides/web/customizing-routes/

test/browser/features/react-router.feature
```diff
@@ -0,0 +1,21 @@
+@skip_chrome_61
```
djskinner, line 1: react router doesn't work in Chrome 61 as AbortController does not exist

test/browser/features/fixtures/packages/rollup.config.mjs
djskinner: Now I'm looking at the vue integration I think it makes sense to move this into its own config for the react router fixture, possibly extending this rollup.config.mjs.

packages/platforms/browser/lib/index.ts
```diff
@@ -1,4 +1,4 @@
-export { DefaultRoutingProvider, default, onSettle } from './browser'
+export { DefaultRoutingProvider, default, onSettle, ReactRouterRoutingProvider } from './browser'
```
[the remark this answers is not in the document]
gareththackeray: I think we probably want the latter.  Or it could be `import { ReactRouterRoutingProvider } from '@bugsnag/browser-performance/routing-providers'` (and likewise for e.g the `VueRoutingProvider`). 
Is there any advantage to that?  (Fewer packages to maintain?)

packages/platforms/browser/lib/routing-provider/react-router-routing-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+  return [
+    ...routes.map(route => `${prefix}${route.path || ''}`),
+    ...routes.reduce<string[]>(
+      (accum, route) => [...accum, ...(route.children ? flattenRoutes(route.children, `${prefix}${route.path}`) : [])],
+      []
+    )
+  ]
+}
+
+export const createReactRouterRoutingProvider = (onSettle: OnSettle, location: Location) => {
```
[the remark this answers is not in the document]
djskinner: For react-router `createBrowserRouter`, yes. You need to pass in all your routes when you create the router and there is no mechanism to change it. Vue router has a concept of dynamic routes, which is why in that PR I've opted to accept the router itself, so `router.getRoutes()` can be called to avail ourselves of any newly added (or removed) routes since the router was initially created.

packages/platforms/browser/lib/routing-provider/react-router-routing-provider.ts
djskinner: I just noticed the default route resolve defaults to `/`:

```
export const defaultRouteResolver: RouteResolver = (url: URL) => url.pathname || '/'
```

gareththackeray: That's just to ensure we send at least `/` even on a root URL (with no path). It's a special case because we just take the full path in that case.

In the case under discussion I would assume that `''` is an invalid route?  But perhaps we should look for the result of `.match()` returning `null` and default that to `no-route-found`.  If it is (somehow) `''` then we could convert that to `'/'`, for the same reason as in the default.

packages/react-router/lib/react-router-routing-provider.ts
```diff
@@ -0,0 +1,31 @@
...
+    ...routes.map(route => `${prefix}${route.path || ''}`),
+    ...routes.reduce<string[]>(
+      (accum, route) => [...accum, ...(route.children ? flattenRoutes(route.children, `${prefix}${route.path}`) : [])],
+      []
+    )
+  ]
+}
+
+export const createReactRouterRoutingProvider = (onSettle: OnSettle, location: Location) => {
+  const DefaultRoutingProvider = createDefaultRoutingProvider(onSettle, location)
```
gingerbenw, line 21: do you need to re-create a default routing provider here, or can you just import `DefaultRoutingProvider` from browser and extend it?

yousif-bugsnag (approved):
LGTM

gingerbenw (approved):

packages/react-router/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "type": "git",
+    "url": "git+https://github.com/bugsnag/bugsnag-js-performance.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "build": "npm run build:npm && npm run build:cdn",
+    "build:npm": "rollup --config rollup.config.npm.mjs",
+    "build:cdn": "echo 1",
```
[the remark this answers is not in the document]
djskinner: I can't remember why I added this in the first place but I noticed when doing the Vue it doesn't seem necessary there, so will remove.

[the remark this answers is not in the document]
djskinner: Oh yeah its cause the CI build fails without it

```
@bugsnag/react-router-performance: npm ERR! Lifecycle script `build:cdn` failed with error: 
@bugsnag/react-router-performance: npm ERR! Error: Missing script: "build:cdn"
```

PR #286, opened 2023-08-07
update .eslintignore

djskinner:


.eslintignore
```diff
@@ -1,4 +1,4 @@
-dist
+build
```
gingerbenw, line 1: `dist` is still used by the test fixtures when you have debugging set to true.

```suggestion
dist
build
```

gingerbenw (dismissed):
need to re-add `dist` to the ignore file

tomlongridge (approved):

PR #289, opened 2023-08-09
build: remove package-lock file

tomlongridge:
## Goal

Checking in the `package-lock.json` file means that CI and Dev environments have a consistent set of dependencies. However as they aren't published to npm, this means that users of our package might have a different experience.

This PR removes the lockfile, making necessary changes to the build files, and bumps a couple of dependencies to remove the reported vulnerabilities, albeit all in the development dependencies.

## Changeset

* `.github/workflows` the `setup-node` action requires a lockfile if the cache is used, so I've taken the cache setting off so dependencies are no longer cached
* Various build scripts – amended to perform a `npm i` throughout rather than a `ci` install, again as that's for lockfiles
* `package.json`:
  * bumped lerna up a major version (6 -> 7) and removed deprecated `useWorkspaces` option, which now is not required as it will use the `workspaces` setting in the `package.json`
  * used `overrides` to set `semver` to `~7.5.3` as `eslint-plugin-compat` currently pulls in a version with a [vulnerability](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw)

## Testing

* Relying on CI
* todo: we need to test lerna on the latest major version


gingerbenw (commented):
my heart (and the [npm documentation](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json), and the [yarn documentation](https://classic.yarnpkg.com/en/docs/yarn-lock)) tells me "commit your lockfiles" but I can see the argument for removing them as they don't get published. personally, I would suggest updating the dependencies more frequently, as our test suite should catch any breakages.

PR #295, opened 2023-08-21
Add e2e test pipeline for React Native

yousif-bugsnag:
## Goal

Adds the basic e2e test setup for React Native performance

## Design

- The test fixture is generated by running the script `.bin/generate-react-native-fixture` - this script calls `react-native init` using the React Native version specified by an environment variable `RN_VERSION` and outputs the fixture to `build/react-native/<RN_VERSION>` - at the moment tests are run against RN version 0.72
- The script will generate an Android APK and/or iOS ipa if the respective `BUILD_ANDROID` and `BUILD_IOS` environment variables are set
- Scenarios are contained in the `test/react-native/features/fixtures/performance-scenarios` package - this is packed and installed into the test fixture by the above script
- The test fixture uses the `react-native-file-access` package to read the `fixture_config.json` file that is pushed to the device by Appium
	- on Android, maze runner's config file directory is overridden to `/data/local/tmp` because the default `/sdcard/...` directory is inaccessible on some devices (see `env.rb` and `ConfigFileReader.js`)
- Scenarios are loaded via maze runner commands which the test fixture polls for when launched. 
	- Each scenario is implemented as a 'root' app component; when a command is received by the test fixture the root app component is replaced with the specified scenario (see `ScenarioLauncher.js`)

#### TODO:
- Currently the fixture doesn't actually install the `@bugsnag/react-native/performance` package- this will be added in a subsequent PR once  #292 is merged
- The test fixture itself can be built and run locally as well as on CI, but at the moment scenarios can only be launched via maze runner commands. At some point we should add an alternative mechanism for launching scenarios manually so that they can be run and debugged locally

## Testing

Added a test scenario and a feature file that verifies the scenario component is launched by the test fixture


bin/generate-react-native-fixture
```diff
@@ -0,0 +1,90 @@
...
+  process.exit(1)
+}
+
+const rnVersion = process.env.RN_VERSION
+const fixtureDir = resolve(__dirname, '../build/react-native/', rnVersion)
+
+if (!process.env.SKIP_GENERATE_FIXTURE) {
+    // remove the fixture directory if it already exists
+  if (fs.existsSync(fixtureDir)) {
+    execSync(`rm -rf ${fixtureDir}`, { stdio: 'inherit' })
```
gingerbenw, line 18: I think you can fix the code scanning issue by using fs

```suggestion
    fs.rmdir(fixtureDir, { recursive: true, force: true })
```

bin/generate-react-native-fixture
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  // create the test fixture
+  execSync(
+    `npx react-native init reactnative --package-name com.bugsnag.fixtures.reactnative.performance --directory ${fixtureDir} --version ${rnVersion} --npm --skip-install`,
```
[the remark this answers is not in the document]
gingerbenw: not sure we can do much about these to be honest...

test/react-native/features/fixtures/app/ios/exportOptions.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>teamID</key>
+    <string>7W9PZ27Y5F</string>
```
gingerbenw, line 6: I take it this is safe to leave here?

yousif-bugsnag: I assumed so, since this was lifted from the test fixtures in the notifier repo

PR #298, opened 2023-08-22
[PLAT-10654] Generate and persist device ID

gingerbenw:
## Goal

Generate and persist device IDs using `@bugsnag/cuid` and `async-storage`

## Changeset

- Move `cuid` ts declaration to root and share between required ts configs
- Move persisted payload validation to core package to share between browser and react native
- Add config option to react-native
- Add persistence to react-native using `async-storage` falling back to in-memory persistence when not available

## Testing

Test `device.id` attribute matches the expected regex in resource attribute source unit tests

### DIFF
--- a/packages/platforms/react-native/lib/persistence.ts
+++ b/packages/platforms/react-native/lib/persistence.ts
@@ -5,17 +5,12 @@ import {
   toPersistedPayload,
   InMemoryPersistence
 } from '@bugsnag/core-performance'
-import type { AsyncStorageStatic } from '@react-native-async-storage/async-storage'
+import AsyncStorage, { type AsyncStorageStatic } from '@react-native-async-storage/async-storage'
 
 export function getReactNativePersistence (): Persistence {
   // use @react-native-async-storage/async-storage if it's installed
   try {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const AsyncStorage = require('@react-native-async-storage/async-storage')
-
-    if (AsyncStorage) {
-      return new ReactNativePersistence(AsyncStorage)
-    }
+    if (AsyncStorage) return new ReactNativePersistence(AsyncStorage)
   } catch {}
 
   // store items in memory if @react-native-async-storage/async-storage isn't available
--- a/packages/platforms/react-native/lib/persistence.ts
+++ b/packages/platforms/react-native/lib/persistence.ts
@@ -1,19 +1,28 @@
 import {
+  InMemoryPersistence,
   toPersistedPayload,
   type Persistence,
   type PersistenceKey,
   type PersistencePayloadMap
 } from '@bugsnag/core-performance'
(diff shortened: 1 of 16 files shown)


packages/platforms/react-native/lib/persistence.ts
```diff
@@ -0,0 +1,44 @@
...
+  type Persistence,
+  type PersistenceKey,
+  type PersistencePayloadMap,
+  toPersistedPayload,
+  InMemoryPersistence
+} from '@bugsnag/core-performance'
+import type { AsyncStorageStatic } from '@react-native-async-storage/async-storage'
+
+export function getReactNativePersistence (): Persistence {
+  // accessing localStorage can throw on some browsers, so we have to catch
```
yousif-bugsnag, line 11: browser-specific comment?

yousif-bugsnag (approved):

packages/platforms/react-native/lib/persistence.ts
gingerbenw: 🔥

packages/platforms/react-native/tests/resource-attributes-source.test.ts
```diff
@@ -6,56 +6,24 @@ describe('resourceAttributesSource', () => {
...
-    ]))
+    function getAttribute (key: string) {
+      const matchingAttribute = jsonAttributes.find(attribute => attribute?.key === key)
+      return matchingAttribute?.value
+    }
+
+    expect(getAttribute('bugsnag.app.code_bundle_id')).toStrictEqual({ stringValue: '12345678' })
+    expect(getAttribute('deployment.environment')).toStrictEqual({ stringValue: 'test' })
+    expect(getAttribute('device.id')).toStrictEqual({ stringValue: expect.stringMatching(/^c[a-z0-9]{20,32}$/) })
+    expect(getAttribute('device.manufacturer')).toStrictEqual({ stringValue: 'Apple' })
```
gingerbenw, line 19: Changed these to an individual assertion each so they're easier to track down when something goes wrong.

packages/platforms/react-native/lib/persistence.ts
```diff
@@ -0,0 +1,45 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const asyncStorage = require('@react-native-async-storage/async-storage')
+    if (isAsyncStorage(asyncStorage)) {
+      return new ReactNativePersistence(asyncStorage)
+    }
+  } catch {}
+
+  // store items in memory if @react-native-async-storage/async-storage isn't available
+  return new InMemoryPersistence()
+}
```
gingerbenw, line 25: @yousif-bugsnag  updated! this now checks if asnycStorage is present and has the API we need, otherwise falls back to using InMemoryPersistence! tested on iOS simulator and working as anticipated.

yousif-bugsnag (approved):

PR #301, opened 2023-08-25
Prevent React Native clock from sticking

gingerbenw:
## Goal

Remove the logic that prevents the clock from moving backwards in react native, if the clock were to drift back by a significant amount (say, from changing timezone) - every span in the future would keep using exactly the same timestamp.


gingerbenw:
Raising a separate PR to look at the odd code coverage values.

PR #304, opened 2023-08-31
Add wrapperComponentProvider config option (React Native)

yousif-bugsnag:
## Goal

React native's `AppRegistry` only permits one `WrapperComponentProvider` per app and there isn't a way to retrieve an existing one.

This PR adds a new configuration option named `wrapperComponentProvider` which allows users to provide their own `WrapperComponentProvider` if they are already using one. 

If provided, the app start wrapper component will be used to wrap the user-provided wrapper component, so that the app start instrumentation wrapper is still at the root.

## Testing

This is difficult to unit test effectively - relies instead on a new e2e test scenario that asserts both that the the app start is instrumented and the supplied wrapper component is rendered as expected


gingerbenw (approved):
checking for a rendered item within the provided wrapper component seems a sensible approach to me 👍🏻

PR #309, opened 2023-09-07
Generate ids by only iterating over a single array

gingerbenw:
## Goal

Improve performance

### DIFF
--- a/packages/platforms/react-native/lib/id-generator.ts
+++ b/packages/platforms/react-native/lib/id-generator.ts
@@ -15,12 +15,13 @@ const idGenerator: IdGenerator = {
   generate (bits: BitLength): string {
     const bytes = new Uint8Array(bits / 8)
 
-    const randomValues = []
+    let randomValue = ''
+
     for (let i = 0; i < bytes.length; i++) {
-      randomValues.push((Math.random() * 255) | 0)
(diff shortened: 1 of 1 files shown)


packages/platforms/react-native/lib/id-generator.ts
```diff
@@ -15,12 +15,13 @@ const idGenerator: IdGenerator = {
   generate (bits: BitLength): string {
     const bytes = new Uint8Array(bits / 8)
```
imjoehaines, line 16: This `Uint8Array` is also unnecessary since we only use it for the `length`, which is always the value we pass to the constructor:

```js
> const bits = 128
>
> new Uint8Array(bits / 8).length
// => 16
> bits / 8
// => 16
```

```suggestion
    const bytes = bits / 8
```

packages/platforms/react-native/lib/id-generator.ts
```diff
@@ -15,12 +15,13 @@ const idGenerator: IdGenerator = {
...
 
-    const randomValues = []
+    let randomValue = ''
+
     for (let i = 0; i < bytes.length; i++) {
```
imjoehaines, line 20:
```suggestion
    for (let i = 0; i < bytes; i++) {
```

imjoehaines (approved):

PR #317, opened 2023-09-12
upload react-router-performance to CDN on release

djskinner:
## Goal

We agreed to make the `@bugsnag/react-router-performance` CDN bundle available on the CDN

## Design

Same approach as for `@bugsnag/browser-performance`

## Changeset

Include react-router-performance bundles in the upload script

## Testing

Will need testing upon release

### DIFF
--- a/bin/cdn-upload
+++ b/bin/cdn-upload
@@ -24,7 +24,11 @@ const files = [
   'bugsnag-performance.js',
   'bugsnag-performance.js.map',
   'bugsnag-performance.min.js',
-  'bugsnag-performance.min.js.map'
+  'bugsnag-performance.min.js.map',
+  'bugsnag-react-router-performance.js',
+  'bugsnag-react-router-performance.js.map',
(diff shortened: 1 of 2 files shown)


imjoehaines:
The release script needs to change as it will only check if the browser package has been updated before uploading to the CDN:

https://github.com/bugsnag/bugsnag-js-performance/blob/9a0eca7f223093bfa84a213d6220f939c333553c/bin/release.sh#L31-L32

The CDN upload script should probably then take the package and files to upload as arguments, so we don't need to upload everything all at once. Otherwise we'll bust the cache unnecessarily. e.g.:

```sh
# upload the browser package
$ ./bin/cdn-upload packages/platforms/browser/ build/bugsnag-performance.*
# upload the react router package
$ ./bin/cdn-upload packages/react-router build/bugsnag-react-router-performance.*
```

imjoehaines:
Also I'd say this shouldn't block the release — we can upload to the CDN separately if necessary to get the release out earlier

PR #321, opened 2023-09-15
Network spans for React Native

yousif-bugsnag:
## Goal

Adds network instrumentation for fetch and xhr requests to the React Native library

## Design

To support this the request trackers from browser performance have been moved into a common package that is now used in both browsers and React Native.

Each platform still has its own separate network request plugin for a couple of reasons:
- React Native's fetch implementation uses a polyfill that uses XMLHttpRequest under the hood, so the react native plugin uses a single xhr request tracker to track both fetch and xhr requests.
- network spans on react native will eventually have additional attributes for connection type and subtype (not covered in this PR)

The `networkRequestCallback` implementations have also been kept separate as the types are incompatible - in the browser the callback is also used for resource load spans and so the `type` property could be any value returned from `PerformanceResourceTiming.initiatorType`, whereas on React Native the callback's type will always be `xmlhttprequest`


packages/request-tracker/lib/request-tracker-xhr.ts
```diff
@@ -14,21 +14,21 @@ interface RequestData {
 
 type ReadyStateChangeHandler = (this: XMLHttpRequest, ev: Event) => any
 
-function createXmlHttpRequestTracker (window: WindowWithXmlHttpRequest, clock: Clock): RequestTracker {
+function createXmlHttpRequestTracker (global: GlobalWithXmlHttpRequest, clock: Clock): RequestTracker {
```
imjoehaines, line 17: This should probably take `XMLHttpRequest` and `document?` separately — the fetch version needs a global object because it overwrites `global.fetch` directly, but this version can work from only the `XMLHttpRequest` object as it overwrites the prototype

It's pretty confusing for `react-native/lib/client.ts` to pass a new object with XHR — it looks like that should never work because overwriting XHR on that object would do nothing

test/react-native/features/fixtures/scenario-launcher/scenarios/NetworkRequestCallbackScenario.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // modify the url for the fetch request
+    return { ...networkRequestInfo, url: networkRequestInfo.url + '&not-your-ordinary-url=true' }
+  }
+}
+
+const fetchSuccess = async () => {
+  try {
+    await fetch('https://google.com/?fetch=true')
```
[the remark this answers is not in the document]
gingerbenw: opted for `bugsnag.com` as we don't know the exact maze runner address, leading to having to check against regex strings in the attributes assertion. I was also seeing issues with invalid URIs when trying to use the maze runner address.

packages/platforms/react-native/lib/network-request-callback.ts
imjoehaines, line 1: This is almost an exact duplicate of [`network-request-callback.ts` in the browser package](https://github.com/bugsnag/bugsnag-js-performance/blob/eb9c74ef7ee136eb46a16922db4013024cb0dbb0/packages/platforms/browser/lib/network-request-callback.ts)

Is that intentional because they might diverge in future or can we unify them (bar `NetworkRequestInfo.type`)?

packages/request-tracker/rollup.config.mjs
```diff
@@ -0,0 +1,4 @@
+import fs from 'fs'
```
imjoehaines, line 1:
```suggestion
```

packages/request-tracker/tests/request-tracker-xhr.test.ts
imjoehaines, line 1: Could use a test that proves `document` is optional

packages/request-tracker/tests/url-helpers.test.ts
imjoehaines, line 1: Could use a test that proves `baseUrl` is optional

test/react-native/features/fixtures/scenario-launcher/scenarios/NetworkRequestCallbackScenario.js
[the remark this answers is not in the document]
gingerbenw: We don't necessarily know the order of the spans since the other suggested changes - do you think it's better to change it back and use a maze-runner endpoint or is hitting bugsnag.com good enough™?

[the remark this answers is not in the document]
imjoehaines: Something like this should do:

```ruby
Then('a span named {string} has an attribute {string} matching the regex {string}') do |span_name, attribute_name, regex|
  spans = spans_from_request_list(Maze::Server.list_for('traces'))
  named_spans = spans.find_all { |span| span['name'] == span_name }

  Maze.check.false(named_spans.empty?)
  Maze.check.true(named_spans.any? do |span|
    span['attributes'].any? do |span_attribute|
      span_attribute['key'] == attribute_name && span_attribute['value']['stringValue'].match?(regex)
    end
  end)
end
```

Avoiding the internet would be good as it's a big source of possible flakes and if we can't reach Maze Runner then the tests would fail anyway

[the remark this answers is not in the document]
gingerbenw: Raised PLAT-10956 to cover this change.

gingerbenw (approved):

PR #334, opened 2023-09-28
[PLAT-10863] Network type attribute

gingerbenw:


packages/platforms/react-native/lib/span-attributes-source.ts
```diff
@@ -1,14 +1,25 @@
 import { type SpanAttributesSource } from '@bugsnag/core-performance'
-import { type ReactNativeConfiguration } from './config'
+import NetInfo from '@react-native-community/netinfo'
 import { type AppStateStatic } from 'react-native'
+import { type ReactNativeConfiguration } from './config'
+import getNetworkConnectionType from './get-network-connection-type'
 
 export function createSpanAttributesSource (appState: AppStateStatic) {
+  let connectionType = 'unknown' // TODO: do we need to set an initial value or will the event listener fire in the first instance?
```
yousif-bugsnag, line 8: I think this TODO can be removed because NetInfo should call the event listener with the current state on subscribe

packages/platforms/react-native/lib/span-attributes-source.ts
```diff
@@ -1,14 +1,25 @@
...
+
+  // Subscribe to network changes
+  NetInfo.addEventListener(state => {
+    connectionType = getNetworkConnectionType(state.type)
+    console.log(`[BusgnagPerformance] connection state: ${JSON.stringify(state)}`)
```
yousif-bugsnag, line 13: this `console.log` call needs removing

yousif-bugsnag (dismissed):

packages/platforms/react-native/package.json
```diff
@@ -18,7 +18,9 @@
     "@bugsnag/core-performance": "^1.2.0",
     "@bugsnag/cuid": "^3.0.2",
     "@bugsnag/delivery-fetch-performance": "^1.2.0",
-    "@bugsnag/request-tracker-performance": "^1.2.0"
+    "@bugsnag/request-tracker-performance": "^1.2.0",
+    "@react-native-community/netinfo": "^9.4.1",
+    "react-native-file-access": "^3.0.4"
```
yousif-bugsnag, line 23: Should these be peer dependencies?

yousif-bugsnag (approved):

bin/generate-react-native-fixture
```diff
@@ -103,6 +103,9 @@ if (!process.env.SKIP_GENERATE_FIXTURE) {
 
   // install react-native-dotenv
   execSync('npm install --save react-native-dotenv', { cwd: fixtureDir, stdio: 'inherit' })
+  
+  // install @react-native-community/netinfo
+  execSync('npm install --save  @react-native-community/netinfo', { cwd: fixtureDir, stdio: 'inherit' })
```
imjoehaines, line 108: Can we do these in one `npm install` command? It's usually much faster that way

PR #337, opened 2023-10-02
allow / as a basename in react-router and vue-router providers

djskinner:
## Goal

It seems reasonable that `'/'` should be treated as a equivalent to `undefined` or `''` as the basename parameter.

## Testing

- Unit test coverage


gingerbenw:
is this expected? "/" is invalid as a base URL using the URL constructor so why do we accept that?

djskinner:
> is this expected behaviour? as far as I know "/" is invalid as a base URL when using the URL constructor so do we need to accept it?

Well `/my/app` is also invalid as a base URL when using the URL constructor but is perfectly valid as a `base` param in vue-router or as a `basename` param in react-router. Whilst it has similar naming I don't think this is really related to constructing URLs with `new URL` where the `base` parameter, if supplied, must be an absolute URL.

`/` is allowed for the `base` and `basename` parameters of vue-router and react-router respectively and is equivalent to using `undefined`. So the thinking is if you specify a `base` parameter of `/` in your vue-router application and we're asking for the same parameter then you should also be able to pass that same value to the routing provider and get the same result. 

Without this change if you legitimately specified `/` as that parameter to your router and also passed the same value to the BugSnag routing provider then no routes will match. So the change here is that `/` and `undefined` are also treated equivalently by our code as they are by the routers.

gingerbenw:
right! okay, understood - thanks. in which case this seems totally fine :)

gingerbenw (approved):

PR #338, opened 2023-10-03
Read device ID from native SDK if present

imjoehaines:
## Goal

Read the device ID from the native SDKs if they are present so that they line up across spans from both JS and native

### DIFF
--- a/packages/platforms/react-native/lib/persistence/index.ts
+++ b/packages/platforms/react-native/lib/persistence/index.ts
@@ -1,12 +1,29 @@
-import File from './file'
+import { File, ReadOnlyFile, NullFile } from './file'
 import FileBasedPersistence from './file-based'
+import NativeBugsnagPerformance from '../native'
+import { Platform } from 'react-native'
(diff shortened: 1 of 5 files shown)


packages/platforms/react-native/lib/persistence/index.ts
```diff
@@ -1,12 +1,29 @@
...
 import FileBasedPersistence from './file-based'
+import NativeBugsnagPerformance from '../native'
+import { Platform } from 'react-native'
 import { Dirs, type FileSystem } from 'react-native-file-access'
 
 const PERSISTED_STATE_VERSION = 1
 const PERSISTED_STATE_PATH = `${Dirs.CacheDir}/bugsnag-performance-react-native/v${PERSISTED_STATE_VERSION}/persisted-state.json`
 
 export default function persistenceFactory (fileSystem: typeof FileSystem): FileBasedPersistence {
+  const nativeDeviceIdFilePath = Platform.select({
```
gingerbenw, line 11: I've come to really like this `Platform.select` API 👌🏻

gingerbenw (approved):

PR #344, opened 2023-10-04
Make react-native-file-access a peer dependency

gingerbenw:
## Goal

`react-native-file-access` requires linking and must be installed directly by the user - this PR changes it from a dependency to a peerDependency


lemnik (approved):
LGTM

PR #352, opened 2023-10-13
[PLAT-10765] Fix PerformanceObserver.entryTypes is required

gingerbenw:
## Goal

handle a known bug in certain browsers, where the `entryType` parameter should be optional in `PerformanceObserver.observe()`

https://bugs.webkit.org/show_bug.cgi?id=209216

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
@@ -101,6 +101,11 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
       }
     })
 
-    observer.observe({ type: 'resource', buffered: true })
+    try {
+      observer.observe({ type: 'resource', buffered: true })
+    } catch (err) {
+      configuration.logger.warn('Unable to get previous resource loads as buffered observer not supported, only showing resource loads from this point on')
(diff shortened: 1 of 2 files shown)


packages/platforms/browser/lib/auto-instrumentation/resource-load-plugin.ts
```diff
@@ -101,6 +101,11 @@ export class ResourceLoadPlugin implements Plugin<BrowserConfiguration> {
       }
     })
 
-    observer.observe({ type: 'resource', buffered: true })
+    try {
+      observer.observe({ type: 'resource', buffered: true })
+    } catch (err) {
+      configuration.logger.warn('Unable to get previous resource loads as buffered observer not supported, only showing resource loads from this point on')
+      observer.observe({ entryTypes: ['resource'] })
```
gareththackeray, line 108: Assume `observer.observe({ entryTypes: ['resource'], buffered: true })` doesn't work?

gingerbenw: no, it's quite odd! `entryTypes` can only exist on it's own.

gingerbenw: it does not play well with others.

PR #354, opened 2023-10-13
Add `RetryQueueDirectory`

imjoehaines:
## Goal

This PR adds an abstraction over a directory for use with a new file-based retry queue to be added in a future PR

This is supported by a new fake implementation of `react-native-file-access` as the mock that ships with the library doesn't really deal with directories

The new `FileSystemFake` is implemented in 8919aaddacdac13fb15a4ea036554980d0a4261c, which is then consumed by the unit tests in the new `RetryQueueDirectory` in 202c62cb9c3fcce92a2768cc31cfa1074527bf9a

### DIFF
--- a/packages/platforms/react-native/tests/utilities/file-system-fake.ts
+++ b/packages/platforms/react-native/tests/utilities/file-system-fake.ts
@@ -6,11 +6,6 @@ type FileSystemEntry
 
 type FileSystem = Map<string, FileSystemEntry>
 
-// TODO: keep?
-// function basename (path: string): string {
-//   return path.split('/').pop() || '/'
-// }
-
 function dirname (path: string): string {
   return path.split('/').slice(0, -1).join('/')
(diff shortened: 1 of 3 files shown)


packages/platforms/react-native/tests/utilities/file-system-fake.ts
```diff
@@ -0,0 +1,110 @@
...
+type FileSystemEntry
+  = { type: FileSystemType.File, name: string, data: string }
+  | { type: FileSystemType.Directory, name: string }
+
+type FileSystem = Map<string, FileSystemEntry>
+
+// TODO: keep?
+// function basename (path: string): string {
+//   return path.split('/').pop() || '/'
+// }
```
gingerbenw, line 12: probably don't need to keep then! :)

gingerbenw (approved):

PR #355, opened 2023-10-16
Increase maximum spans in react-router test

gingerbenw:
## Goal

Reduce number of flakes in end to end tests

### DIFF
--- a/test/browser/features/fixtures/packages/react-router/src/index.jsx
+++ b/test/browser/features/fixtures/packages/react-router/src/index.jsx
@@ -52,7 +52,7 @@ const router = createBrowserRouter(routes, { basename })
 BugsnagPerformance.start({
     apiKey,
     endpoint,
-    maximumBatchSize: 13,
+    maximumBatchSize: 14,
     batchInactivityTimeoutMs: 5000,
     autoInstrumentNetworkRequests: false,
     routingProvider: new ReactRouterRoutingProvider(routes, basename)


imjoehaines (commented):
I'm not sure about this — the page load span is relevant since `resolveRoute` is used to get the route name attribute

What is it about the 13 spans that's causing flakes? Is the number of spans variable?

gingerbenw:
> I'm not sure about this — the page load span is relevant since `resolveRoute` is used to get the route name attribute
> 
> What is it about the 13 spans that's causing flakes? Is the number of spans variable?

Yeah, we had a span overflow to the next payload which caused a failure. We could just increase the maximum number of spans to 14, but I didn't feel like the full page load wasn't relevant to the route change feature.

imjoehaines:
> > I'm not sure about this — the page load span is relevant since `resolveRoute` is used to get the route name attribute
> > What is it about the 13 spans that's causing flakes? Is the number of spans variable?
> 
> Yeah, we had a span overflow to the next payload which caused a failure. We could just increase the maximum number of spans to 14, but I didn't feel like the full page load was relevant to the route change feature.

It isn't just about route changes though, it's also relevant to full page loads because of the `resolveRoute` call here:

https://github.com/bugsnag/bugsnag-js-performance/blob/bc5592859400612a200347f39baf3ef1e9a8e2bf/packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts#L67

imjoehaines (approved):

PR #357, opened 2023-10-17
Add file-based retry queue for React Native

imjoehaines:
## Goal

This PR adds a file based retry queue that works by writing failed payloads to disk and reading from them when `flush` is called

If delivery is successful or fails with the `failure-discard` state, the file gets deleted. Otherwise the file is left on disk to be retried on the next `flush` call

Files are stored with a filename in the format `retry-<timestamp>-<span ID>.json` where `timestamp` is the largest `endTimeUnixNano` in the payload's spans

Any file with a timestamp more than 24 hours in the past or future will be deleted when `flush` is called. Files with invalid names (i.e. do not match the format) will also be deleted

## Testing

This has only been unit tested so far so is not actually being used by the SDK yet

### DIFF
--- a/packages/platforms/react-native/lib/retry-queue/file-based.ts
+++ b/packages/platforms/react-native/lib/retry-queue/file-based.ts
@@ -73,6 +73,7 @@ const enum FlushOutcome { DeleteFile, LeaveFile }
 export default class FileBasedRetryQueue implements RetryQueue {
   private readonly delivery: Delivery
   private readonly directory: Directory
+  private requestQueue: Promise<void> = Promise.resolve()
 
   constructor (delivery: Delivery, directory: Directory) {
     this.delivery = delivery
@@ -101,18 +102,22 @@ export default class FileBasedRetryQueue implements RetryQueue {
   }
 
   async flush (): Promise<void> {
-    const files = await this.directory.files()
+    this.requestQueue = this.requestQueue.then(async () => {
+      const files = await this.directory.files()
 
-    for (const filename of files) {
-      try {
-        const outcome = await this.flushFile(filename)
+      for (const filename of files) {
+        try {
+          const outcome = await this.flushFile(filename)
 
-        if (outcome === FlushOutcome.DeleteFile) {
-          await this.directory.delete(filename)
+          if (outcome === FlushOutcome.DeleteFile) {
+            await this.directory.delete(filename)
+          }
+        } catch {
         }
-      } catch {
       }
-    }
+    })
+
+    await this.requestQueue
   }
 
   private async flushFile (filename: string): Promise<FlushOutcome> {
--- a/packages/platforms/react-native/lib/retry-queue/file-based.ts
+++ b/packages/platforms/react-native/lib/retry-queue/file-based.ts
@@ -1,0 +1,150 @@
+import {
+  type Delivery,
+  type DeliveryPayload,
+  type DeliverySpan,
+  type RetryQueue,
+  type TracePayload
+} from '@bugsnag/core-performance'
+import type Directory from './directory'
+import timestampFromFilename from './timestamp-from-filename'
+
+function getLatestSpan (body: DeliveryPayload): DeliverySpan | undefined {
+  let latestSpan: DeliverySpan | undefined
+  let biggestTimestamp: bigint | undefined
+
+  for (const resourceSpan of body.resourceSpans) {
+    for (const scopeSpan of resourceSpan.scopeSpans) {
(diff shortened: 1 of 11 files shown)


packages/platforms/react-native/lib/retry-queue/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  type RetryQueue,
+  type RetryQueueFactory
+} from '@bugsnag/core-performance'
+
+import RetryQueueDirectory, { type MinimalFileSystem } from './directory'
+import FileBasedRetryQueue from './file-based'
+import { PERSISTENCE_DIRECTORY } from '../persistence'
+
+export default function createRetryQueueFactory (fileSystem: MinimalFileSystem): RetryQueueFactory {
+  return function fileBasedQueueFactory (delivery: Delivery, _retryQueueMaxSize: number): RetryQueue {
```
gingerbenw, line 12: unused `_retryQueueMaxSize` parameter

yousif-bugsnag (approved):
LGTM!

packages/platforms/react-native/lib/retry-queue/file-based.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    // we use the latest span's timestamp in the filename so we can decide
+    // whether to keep or discard the span without parsing the payload
+    // as we can't be sure of nanosecond precision in JS environments, we also
+    // append the span ID so file names are unique across batches
+    const filename = `retry-${span.endTimeUnixNano}-${span.spanId}.json`
+
+    this.directory.write(filename, JSON.stringify(payload))
+      .catch(() => {})
```
gingerbenw, line 96: should we at least log a warning if this fails?

packages/platforms/react-native/tests/retry-queue/file-based.test.ts
```diff
@@ -0,0 +1,350 @@
...
+      await flushPromises()
+
+      expect(await directory.files()).toStrictEqual([
+        `retry-${validEndTime + BigInt(2)}-c.json`,
+        `retry-${validEndTime + BigInt(1)}-b.json`,
+        `retry-${validEndTime}-a.json`
+      ])
+
+      delivery.setNextResponseState('failure-discard')
+      delivery.setNextResponseState('success')
```
gingerbenw, line 241: I'm not quite following this test or the next with regards to the response states... we have 3 payloads to deliver, the first one fails with `failure-discard` and the next one succeeds? so why are all 3 payloads removed? I'm sure I've got the logic wrong here...

imjoehaines: The files are deleted when either:

1. delivery fails permanently (i.e. `failure-discard`) as delivery will never succeed so the payload can only be discarded
2. delivery succeeds as otherwise we'd send the payload again on the next `flush` and get duplicates in the dashboard

packages/platforms/react-native/lib/retry-queue/index.ts
imjoehaines: That's intentional since the `RetryQueueFactory` type has this parameter:

https://github.com/bugsnag/bugsnag-js-performance/blob/0448e7a43823bb49b555eee48bcb497f7e5d203f/packages/core/lib/retry-queue.ts#L8

packages/platforms/react-native/lib/retry-queue/file-based.ts
imjoehaines: @gareththackeray ?

gareththackeray: I would have thought so. Are there reasons why we might not want to?

imjoehaines: Only that you probably couldn't do anything about it and by default they'd just be logged to the end user's device so wouldn't be visible to the developer

packages/platforms/react-native/lib/retry-queue/file-based.ts
```diff
@@ -101,18 +102,22 @@ export default class FileBasedRetryQueue implements RetryQueue {
   }
 
   async flush (): Promise<void> {
-    const files = await this.directory.files()
+    this.requestQueue = this.requestQueue.then(async () => {
```
imjoehaines, line 105: Realised this request queueing was missing so calling `flush` multiple times in quick succession could deliver the same payload multiple times

Now it always waits for the previous `flush` to finish before starting another one

gingerbenw (approved):

PR #362, opened 2023-10-20
Listen for pagehide and pageshow events in backgrounding listener

yousif-bugsnag:
## Goal

On some versions of Mobile Safari the `visibilitychange` event does not fire if a page is cached/suspended when navigating away, which can mean spans that should be discarded are left open.

This PR updates the backgrounding listener to also listen for `pagehide` and `pageshow` events which do fire in that scenario.

The backgrounding listener now tracks the current backgrounding state to avoid notifying subscribers more than once when both `visibilitychange` and `pageshow`/`pagehide` events are fired

### DIFF
--- a/packages/platforms/browser/lib/backgrounding-listener.ts
+++ b/packages/platforms/browser/lib/backgrounding-listener.ts
@@ -1,6 +1,7 @@
 import {
   type BackgroundingListener,
-  type BackgroundingListenerCallback
+  type BackgroundingListenerCallback,
+  type BackgroundingListenerState
 } from '@bugsnag/core-performance'
 
 interface DocumentForVisibilityState {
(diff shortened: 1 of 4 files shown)


packages/platforms/browser/lib/backgrounding-listener.ts
```diff
@@ -1,35 +1,61 @@
...
       callback(state)
     }
+  }
+
+  window.document.addEventListener('visibilitychange', function () {
+    const state = window.document.visibilityState === 'hidden'
+      ? 'in-background'
+      : 'in-foreground'
+
+    notifySubscribers(state)
```
gareththackeray, line 46: Would it be sensible to keep track of the last known state and only notify subscribers when it changes?  To avoid getting multiple notifications for effectively the same event?

yousif-bugsnag: Yes it would!

packages/platforms/browser/tests/backgrounding-listener.test.ts
```diff
@@ -29,50 +29,150 @@ class DocumentStub {
   }
 }
 
+class WindowFake {
+  readonly #eventListeners = new Map<string, () => void>()
```
gingerbenw, line 33: is the # just a naming convention here? no issue with it - I'm just curious as it's been cropping up a little bit lately.

imjoehaines: It's a private class field: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

e.g. this is invalid:

```js
const window = new WindowFake()
window.#eventListeners // throws an error
```

gingerbenw: thanks!

PR #363, opened 2023-10-23
Add 'onStart' callback to core

imjoehaines:
## Goal

This allows clients to register a callback that is called when the user calls 'start' so that platforms can then implement custom logic

'onStart' is called before anything else is setup (e.g. plugins) and is passed  the validated configuration object


imjoehaines:
Realised this doesn't actually work for what we need in RN&hellip;

PR #364, opened 2023-10-26
Round timestamps to the nearest integer

yousif-bugsnag:
## Goal

Browser timing APIs usually return a `DOMHighResTimestamp` (aka a double). These values are usually integers (i.e. whole milliseconds) but it is valid for them to contain a fractional part, representing nanoseconds

Currently we don’t do anything to handle this, which results in our encoded JSON payloads being invalid according to the OTEL spec

This PR updates the millisecond to nanosecond conversion to round nanoseconds to the nearest integer to ensure we always send a valid payload.

## Design

`performance.now` on React Native also returns a double, so went for implementing this in the core `millisecondsToNanoseconds` function as this is used to convert millisecond values by both of the platform-specific clocks

### DIFF
--- a/packages/core/tests/clock.test.ts
+++ b/packages/core/tests/clock.test.ts
@@ -1,0 +1,8 @@
+import { millisecondsToNanoseconds } from '../lib/clock'
+
+describe('millisecondsToNanoseconds', () => {
+  it('rounds to the nearest integer', () => {
+    expect(millisecondsToNanoseconds(123456789.1122345)).toEqual(123456789_112235)
+    expect(millisecondsToNanoseconds(123456789.1122344)).toEqual(123456789_112234)
+  })
+})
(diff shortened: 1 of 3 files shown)


packages/core/tests/clock.test.ts
```diff
@@ -0,0 +1,8 @@
+import { millisecondsToNanoseconds } from '../lib/clock'
+
+describe('millisecondsToNanoseconds', () => {
+  it('rounds to the nearest integer', () => {
```
imjoehaines, line 4: nit:

```suggestion
  it('rounds to the nearest nanosecond', () => {
```

imjoehaines: just to be a bit clearer what we're actually trying to achieve

imjoehaines (approved):

gingerbenw (approved):

gingerbenw (changes requested):
Changelog!

gingerbenw (approved):

PR #365, opened 2023-10-26
Flush the retry queue in more situations on React Native

imjoehaines:
## Goal

Flush the retry queue in more situations on React Native — because there is a persistent queue, it makes sense to flush more often than in the browser SDK

Specifically the queue is flushed:

- when `BugsnagPerformance.start` is called
- when the app is backgrounded or foregrounded
- when the device gains network connectivity


gingerbenw (approved):
LGTM

PR #366, opened 2023-11-01
Add maze runner tests for persistent state in React Native

imjoehaines:
## Goal

Add maze runner tests for persistent state (sampling probability & device ID) in React Native

## Design

This required some changes to the test fixture's command fetching — we need to add the `persistent-data.json` file to the device before `BugsnagPerformance#start`, as this is when we read the file. i.e. it's too late to create the file when the scenario starts


test/react-native/features/fixtures/scenario-launcher/lib/ScenarioLauncher.js
```diff
@@ -39,6 +36,85 @@ export const launchScenario = async (rootTag) => {
...
+    console.error(`[BugsnagPerformance] creating persisted state directory: ${PERSISTED_STATE_DIRECTORY}`)
+    await FileSystem.mkdir(PERSISTED_STATE_DIRECTORY)
+  }
+
+  console.error(`[BugsnagPerformance] writing to: ${PERSISTED_STATE_PATH}`)
+
+  await FileSystem.writeFile(
+    PERSISTED_STATE_PATH,
+    JSON.stringify(contents)
+  )
```
gingerbenw, line 54: does `writeFile` append the contents here? or would calling `setSamplingProbability` overwrite the data from `setDeviceId` and vice versa?

imjoehaines: There's no merging of the contents so if you called both the second would win

gingerbenw (approved):

PR #367, opened 2023-11-07
[PLAT-11178] Fix invalid sampling probability test

gingerbenw:

### DIFF
--- a/test/react-native/features/fixtures/scenario-launcher/lib/ScenarioLauncher.js
+++ b/test/react-native/features/fixtures/scenario-launcher/lib/ScenarioLauncher.js
@@ -105,7 +105,7 @@ export async function launchScenario (rootTag, clearPersistedData = true) {
       return await launchScenario(rootTag, false)
 
     case 'set-invalid-sampling-probability':
-      await setSamplingProbability(0, 12345)
+      await setSamplingProbability('invalid', 12345)
 
       return await launchScenario(rootTag, false)
 


test/react-native/features/fixtures/scenario-launcher/lib/ScenarioLauncher.js
```diff
@@ -105,7 +105,7 @@ export async function launchScenario (rootTag, clearPersistedData = true) {
       return await launchScenario(rootTag, false)
 
     case 'set-invalid-sampling-probability':
-      await setSamplingProbability(0, 12345)
+      await setSamplingProbability('invalid', 12345)
```
imjoehaines, line 108: Was this a bug on BitBar or our end? It seems weird that this flaked because of the time but the "set-sampling-probability-to-0" works — if this time is valid then shouldn't "set-sampling-probability-to-0" be invalid?

Maybe it should really be 2 scenarios? 1 with an invalid value and 1 with an invalid time

gingerbenw: I think two scenarios would be good. 
this is technically a valid time, right? it's just expired?

imjoehaines: I'm still a bit confused why this was flakey but the valid probability wasn't — I'd say the issue was BitBar devices having a clock set to 0 (so they're in 1970 and within 24 hours of `12345`) but then the valid probability using `Date.now` should have expired so that scenario would fail 🤔

gingerbenw: well it does appear to be passing with that new timestamp, and broken out into separate expired and invalid sampling probability scenarios...

imjoehaines (approved):

PR #375, opened 2023-11-21
add angular router instrumentation

djskinner:
## Goal

Adds support for router instrumentation in angular projects

## Design

Usage

```
import BugsnagPerformance from '@bugsnag/browser-performance';
import { AngularRoutingProvider, bugsnagBootstrapper } from '@bugsnag/angular-performance';

BugsnagPerformance.start({
  apiKey,
  autoInstrumentRouteChanges: true,
  routingProvider: new AngularRoutingProvider(),
})

@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    BrowserModule,
    HttpClientModule,
    AppRoutingModule
  ],
  providers: [
    bugsnagBootstrapper,
  ],
  bootstrap: [AppComponent]
})
export class AppModule { }
```


packages/platforms/browser/lib/on-settle/load-event-end-settler.ts
```diff
@@ -52,8 +52,10 @@ class LoadEventEndSettler extends Settler {
 
     if (isPerformanceNavigationTiming(entry)) {
       settledTime = entry.loadEventEnd
-    } else {
+    } else if (performance.timing) {
       settledTime = performance.timing.loadEventEnd - performance.timing.navigationStart
+    } else {
+      settledTime = 0
```
djskinner, line 58: I added this so that I was able to unit test AngularRoutingProvider was setting the span URL after onSettle. This is perhaps not the way we want to handle this but thought I'd leave it here for discussion as we don't elsewhere seem to be testing the actual exported onSettle (we could add similar coverage to vue-router and react-router).

Without this allowing onSettle to trigger in a test fails because window.performance.timing is not defined in jsdom

packages/vue-router/tests/vue-router-routing-provider.test.ts
```diff
@@ -44,7 +45,7 @@ describe('VueRouterRoutingProvider', () => {
         '[FullPageLoad]/some-route',
         { }
       )
-      const startRouteChangeSpan = jest.fn(() => spanFactory.toPublicApi(span))
+      const startRouteChangeSpan: StartRouteChangeCallback = jest.fn(() => spanFactory.toPublicApi(span) as RouteChangeSpan)
```
djskinner, line 48: the span factory returns `Span` but start route change callback now returns `RouteChangeSpan` with the extended `end` signature

packages/angular/tests/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      jest.advanceTimersByTime(51)
+
+      expect(router.events.subscribe).toHaveBeenCalled()
+
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const handler = jest.mocked(router.events.subscribe).mock.calls[0][0]!
+
+      const navStart = new NavigationStart(0, 'url')
+      // eslint-disable-next-line no-useless-call
+      handler.call(undefined, navStart)
```
gingerbenw, line 46: does this effectively mock a navigation event?

packages/platforms/browser/lib/on-settle/load-event-end-settler.ts
gingerbenw: It sounds like an integration test for onSettle might be a good idea - we do mock quite a lot of implementation.

totally out of scope for this PR though :)

packages/angular/tests/index.test.ts
djskinner: Its just checking things are wired up:
- that we call `router.events.subscribe` with a callback handler function
- that when that callback handler function gets called with a NavigationStart event `startRouteChangeSpan` is called
- that when that callback handler function gets called with a NavigationEnd event  the span returned by `routeChangeSpan` has its `end` method called

packages/angular/lib/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    const base = document.baseURI.replace(window.origin, '')
+    return url.pathname.replace(base, '') || '/'
+  }
+
+  listenForRouteChanges (startRouteChangeSpan: StartRouteChangeCallback): void {
+    // wait for the router to be available
+    if (!globalRouterRef) {
+      setTimeout(() => { this.listenForRouteChanges(startRouteChangeSpan) }, 50)
+      return
+    }
```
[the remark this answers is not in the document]
gingerbenw: it would remove the need for this function to recursively call itself, but I can imagine it's probably not a huge saving. I may have worded it badly of course!

packages/platforms/browser/lib/on-settle/load-event-end-settler.ts
imjoehaines: Do we need to add anything at all here? The fallback to use `now` would trigger if `settledTime` is `undefined` so I think we only need the `else if (performance.timing)` bit

test/browser/features/support/skip.rb
```diff
@@ -16,4 +16,12 @@
 
 Before("@skip_chrome_61") do
   skip_this_scenario("Skipping scenario: Not supported") if Maze.config.browser == "chrome_61"
+end
+
+Before("@skip_firefox_60") do
```
imjoehaines, line 21: Is it worth generating these so we don't need to keep adding specific versions here? e.g. something like:

```ruby
["chrome", "firefox", "safari", "edge"].each do |browser|
  1.upto(1_000) do |version|
    Before("@skip_#{browser}_#{version}") do
      skip_this_scenario("Skipping scenario: Not supported") if Maze.config.browser == "#{browser}_#{version}"
    end
  end
end
```

imjoehaines (approved):

PR #380, opened 2023-12-04
Generate random bytes natively for ID generation

yousif-bugsnag:
## Goal

Some duplicate span IDs were observed in production as `Math.random` isn't random enough.

This PR updates the id generation to source randomness from cryptographically secure number generators in the native layer.

## Design

Two new methods `requestEntropy` and `requestEntropyAsync` have been added to the Native Module. These return random bytes as a 2048 character hex string generated using `SecureRandom` on Android and `SecRandomCopyBytes` on iOS. These characters are then used by the ID generator to generate span and trace IDs by randomly selecting characters from the string using `Math.random()`.

If the Native Module is not available for any reason, we fall back to generating a random string in JS using `Math.random`.

When the client is loaded, a synchronous call to `requestEntropy` is made for an initial set of random values. After 1000 calls, `requestEntropyAsync` is called to refresh the random values asynchronously.

As part of this work, also fixed a bug in the iOS Native Module that was preventing it from working on the old architecture (it was missing an `RCT_EXPORT_` macro on the method), and reinstated the native device attribute e2e tests for old arch on iOS.

## Testing

Tested manually, added some more unit tests, and relied on the existing trace validation in e2e tests.


packages/platforms/react-native/lib/entropy-source.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  let nextIndex = 0
+  const nextByte = () => {
+    const value = pool[nextIndex]
+    nextIndex++
+
+    // when the pool is exhausted, trigger an async operation to regenerate.
+    // We also shuffle the current pool so that we don't have to wait for
```
gareththackeray, line 49: If the pool's exhausted why does shuffling it help?

yousif-bugsnag: The call to the native layer is expensive, so shuffling the pool means we don't need to await the async call and can continue to generate IDs in the meantime

gareththackeray: I must be misunderstanding.  Wouldn't that result in using an id you've already used?

yousif-bugsnag: no - it's not a pool of IDs, it's a sequence of random bytes that is used to generate the IDs. Each ID uses either 8 or 16 bytes  (see `id-generator.ts`)

so by shuffling we get a different sequence of bytes while we wait for a new pool

gingerbenw: if it's an asynchronous call, would it be better to rehydrate the pool after a certain number _near_ the limit has been reached? then we wouldn't need to reshuffle the values we've already got, we can just grab the next allocation, and any subsequent call should have a rehydrated source.

yousif-bugsnag: the issue with that is you don't know how long the call will take or how many IDs you will need to generate before it completes

gingerbenw (approved):

packages/platforms/react-native/__mocks__/react-native.ts
```diff
@@ -101,6 +101,14 @@ type DeviceInfo
   | DeviceInfoAndroid
   | Record<string, never>
 
+const createPool = () => {
+  const pool: number[] = []
+  for (let i = 0; i < pool.length; i++) {
```
imjoehaines, line 106: `pool.length` here is always 0 so this loop is never entered — should this function just be `return []` or should it be filling the array?

yousif-bugsnag: yep, that's a bug!

This has changed now anyway - after some discussion we decided to refactor the native implementation to return the random bytes as a 2048 character hex encoded string

packages/platforms/react-native/lib/id-generator.ts
```diff
@@ -11,18 +16,55 @@ function toHex (value: number): string {
...
+      // pick characters from the pool semi-randomly
+      for (let i = 0; i < chars; i++) {
+        const randomIdx = (Math.random() * (randomPool.length - 1)) | 0
+        id += randomPool[randomIdx]
+      }
+
+      numCalls++
+
+      // if the max number of calls has been reached, refresh the pool asynchronously
+      if (numCalls >= MAX_CALLS) {
```
imjoehaines, line 58: Could do a little bit of squishening if you don't hate it:

```suggestion
      // if the max number of calls has been reached, refresh the pool asynchronously
      if (++numCalls >= MAX_CALLS) {
```

packages/platforms/react-native/lib/id-generator.ts
```diff
@@ -11,18 +16,55 @@ function toHex (value: number): string {
...
+  let numCalls = 0
+  const idGenerator: IdGenerator = {
+    generate (bits: BitLength): string {
+      const chars = bits / 4
+      let id = ''
+
+      // pick characters from the pool semi-randomly
+      for (let i = 0; i < chars; i++) {
+        const randomIdx = (Math.random() * (randomPool.length - 1)) | 0
+        id += randomPool[randomIdx]
```
imjoehaines, line 52: nit:

```suggestion
        const randomIndex = (Math.random() * (randomPool.length - 1)) | 0
        id += randomPool[randomIndex]
```

packages/platforms/react-native/lib/id-generator.ts
```diff
@@ -11,18 +16,55 @@ function toHex (value: number): string {
...
+  // initialise the pool synchronously
+  const randomValues = requestEntropy()
+  let randomPool = randomValues.length > 0 ? randomValues : createRandomString()
+
+  const regeneratePool = async () => {
+    const randomValues = await requestEntropyAsync()
+    randomPool = randomValues.length > 0 ? randomValues : createRandomString()
   }
+
+  let numCalls = 0
```
imjoehaines, line 43: nit:

```suggestion
  let numberOfCalls = 0
```

imjoehaines (approved):
lgtm, just some nitpickyness

packages/platforms/react-native/lib/id-generator.ts
yousif-bugsnag: Always in favour of a bit of squishening

PR #383, opened 2023-12-06
Use file based retry queue in React Native SDK

imjoehaines:


gingerbenw (approved):

imjoehaines:
> Should we update the changelog too?

I'm going to merge this PR as is so it doesn't have to go through CI again just for a changlog update, but we can update it on the release PR if you'd like?

gingerbenw:
> We can update it on the release PR if you'd like?

Yeah, I think that would be good to get it as accurate as possible.

PR #384, opened 2023-12-11
Discard React Native spans when the app is in the background

yousif-bugsnag:
## Goal

Registers a backgrounding listener in the React native client so that any open spans are now discarded when the app moves to the background.


packages/platforms/react-native/lib/backgrounding-listener.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (state === newState) return
+
+    state = newState
+    for (const callback of callbacks) {
+      callback(state)
+    }
+  }
+
+  appState.addEventListener('change', (state: AppStateStatus) => {
+    const newState = state === 'active'
```
gingerbenw, line 37: does this need to be `'unknown'` as well?

yousif-bugsnag: I think `unknown` is an initial state, so shouldn't be sent in an event, but it can't hurt 👍🏾

gingerbenw (approved):

PR #388, opened 2023-12-14
[PLAT-11289] Add createNavigationContainer method

gingerbenw:
## Goal

Add `createNavigationContainer` method to wrap the react navigation container

## Changeset

- Added method for wrapping provided component and monkey patching `onStateChange` method.
- Updated jest configuration for component testing

## Testing

Unit tests for route changes


gingerbenw:
not sure why the test coverage changes aren't showing anything... 🤔

packages/platforms/react-native/tests/retry-queue/index.test.ts
```diff
@@ -4,7 +4,6 @@ import FileSystemFake from '../utilities/file-system-fake'
 import { InMemoryDelivery, makePayloadCreator } from '@bugsnag/js-performance-test-utilities'
 import { AppState } from 'react-native'
 
-// eslint-disable-next-line jest/no-mocks-import
```
gingerbenw, line 7: oh yeah, I fixed this somehow!

gingerbenw: or did I un-fix it?

packages/react-navigation/lib/create-navigation-container.tsx
gingerbenw, line 1: this is actually the main new feature

packages/react-navigation/lib/create-navigation-container.tsx
```diff
@@ -0,0 +1,40 @@
+import BugsnagPerformance from "@bugsnag/react-native-performance";
+import { NavigationContainer, NavigationContainerProps, NavigationContainerRef, useNavigationContainerRef } from "@react-navigation/native";
+import React, { forwardRef, useRef } from 'react';
+import { NavigationContextProvider } from "./navigation-context";
+
+export const createNavigationContainer = <P extends {}>(Container = NavigationContainer) =>
+  forwardRef<NavigationContainerRef<P>, NavigationContainerProps>(
+    (props, _ref) => {
```
yousif-bugsnag, line 8: Not sure if this `forwardRef` is required? the passed in `_ref` isn't used anywhere

gingerbenw: We're using `forwardRef` here to access the props of the component provided as a parameter to the function, but I don't think we actually need the ref, as we will generally just be calling this function with the default `NavigationContainer` exported from `@react-navigation/native` so shouldn't have any imperative handling with `refs` before we access it.

yousif-bugsnag (approved):

PR #391, opened 2024-01-03
[PLAT-11288] Create new CompleteNavigation component

gingerbenw:
## Goal

Create `CompleteNavigation` component, to be included in an application to end an open navigation span

## Testing

Unit tests that the appropriate method is called when a component is mounted or unmounted


packages/react-navigation/lib/complete-navigation.tsx
```diff
@@ -0,0 +1,39 @@
...
+  
+  // @ts-expect-error
+  context: React.ContextType<typeof NavigationContext>;
+
+  constructor(props: Props, context: typeof NavigationContext) {
+    super(props);
+
+    // @ts-expect-error
+    context.blockNavigationEnd();
+  }
```
yousif-bugsnag, line 20: I think these typescript errors can be fixed by the following?
https://legacy.reactjs.org/docs/context.html#classcontexttype

```suggestion
export class CompleteNavigation extends React.Component<Props> {
  static contextType = NavigationContext;
  
  declare context: React.ContextType<typeof NavigationContext>;

  constructor(props: Props) {
    super(props);

    this.context.blockNavigationEnd();
  }
```

yousif-bugsnag: ah maybe not, that breaks the unit tests...

packages/react-navigation/test/utils/create-test-client.ts
yousif-bugsnag, line 1: I'm a bit confused as to why we have this in addition to the mocked `@bugsnag/react-native-performance`?

packages/react-navigation/lib/complete-navigation.tsx
gingerbenw: updated!

packages/react-navigation/test/utils/create-test-client.ts
gingerbenw: gone!

packages/test-utilities/lib/create-span.ts
```diff
@@ -15,6 +16,15 @@ export function createSamplingProbability (rawProbability: number): SpanProbabil
   }
 }
 
+export function createSpan (name: string): Span {
```
yousif-bugsnag, line 19: Can we remove the `name` param here as it's not being used?

packages/react-navigation/lib/complete-navigation.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface Props extends PropsWithChildren {
+  on: 'mount' | 'unmount';
+}
+
+/** End the current navigation span when the component is mounted, unmounted or the `on` prop is `true` */
+export class CompleteNavigation extends React.Component<Props> {
+  static contextType = NavigationContext;
+  
+  context!: React.ContextType<typeof NavigationContext>;
```
yousif-bugsnag, line 12:
```suggestion
  declare context: React.ContextType<typeof NavigationContext>;
```

yousif-bugsnag (approved):
couple of minor suggestions, otherwise LGTM

packages/test-utilities/lib/create-span.ts
gingerbenw: It will get used in the next PR to make assertions on spans a little easier

PR #392, opened 2024-01-10
[PLAT-11284] Cancel span on navigate

gingerbenw:
## Goal

Ensure that opening a new navigation span will cancel an existing one

## Design

Make use of the existing span context features to discard spans with an invalid end time 

## Testing

Update existing unit tests to assert navigation spans are only created when the route changes
Add new unit test to assert open navigation span is ended with an invalid time


packages/react-navigation/lib/navigation-context.tsx
```diff
@@ -63,7 +65,14 @@ export class NavigationContextProvider extends React.Component<Props, State> {
...
     const { currentRoute, client } = this.props;
 
-    if (currentRoute && currentRoute !== this.state.previousRoute) {
+    if (currentRoute && currentRoute !== prevProps.currentRoute) {
+
+      // If there is already an active navigation span, end it with an 
+      // invalid time to cause it to be discarded from the context stack.
+      if (this.currentSpan) {
+        this.currentSpan.end(DISCARDED);
+      }
```
yousif-bugsnag, line 74: I don't think ending the span with an end time of -1 will actually 'discard' the span? i.e. it will get popped off the context stack because it's been ended, but I think it will still get processed and sent...

yousif-bugsnag (approved):

PR #393, opened 2024-01-16
[PLAT-11285] Add Navigation span attributes

gingerbenw:
## Goal

To add required attributes for navigation spans

## Design

As the public `startNavigationSpan` method returns a span without any way to add attributes, this PR refactors the react navigation plugin to accept the span factory as part of it's configuration method, using all internal functionality to start and end spans, and add new attributes.

## Testing

Update existing unit tests to add assertions for new span attributes


packages/core/lib/core.ts
```diff
@@ -101,11 +101,11 @@ export function createClient<S extends CoreSchema, C extends Configuration, T> (
       })
 
       for (const plugin of configuration.plugins) {
-        plugins.push(plugin)
+        plugins.push(plugin as unknown as Plugin<C>)
```
gingerbenw, line 104: little bit of a hack here I would like to see improved.

packages/react-navigation/test/navigation-context.test.tsx
```diff
@@ -1,163 +1,150 @@
...
-import { InMemoryDelivery, VALID_API_KEY, createTestClient } from '@bugsnag/js-performance-test-utilities'
-import type BugsnagPerformance from '@bugsnag/react-native-performance'
+import { type SpanFactory } from '@bugsnag/core-performance'
+import { MockSpanFactory } from '@bugsnag/js-performance-test-utilities'
+import { type ReactNativeConfiguration } from '@bugsnag/react-native-performance'
 import { fireEvent, render, screen } from '@testing-library/react-native'
 import React, { useContext } from 'react'
 import { Button, View } from 'react-native'
 import { NavigationContext, NavigationContextProvider } from '../lib/navigation-context'
-import reactNativePlatformExtensions from './utils/react-native-platform-extensions'
```
yousif-bugsnag, line 8: Can this file be deleted now it's no longer being used?

packages/react-navigation/test/navigation-context.test.tsx
```diff
@@ -1,163 +1,150 @@
-import { type Span } from '@bugsnag/core-performance'
-import { InMemoryDelivery, VALID_API_KEY, createTestClient } from '@bugsnag/js-performance-test-utilities'
-import type BugsnagPerformance from '@bugsnag/react-native-performance'
```
yousif-bugsnag, line 3: Do we still need a `BugsnagPerformance` mock or can that be removed now?

gingerbenw: 💥

packages/react-navigation/test/navigation-context.test.tsx
gingerbenw: 💥

yousif-bugsnag (approved):

PR #404, opened 2024-02-12
[PLAT-11564] Create react-native-navigation-performance package

gingerbenw:
## Goal

Create new `react-native-navigation-performance` package in the monorepo with unit testing configuration


lemnik (approved):
LGTM

PR #409, opened 2024-02-21
[PLAT-11566] Instrument navigation spans

gingerbenw:
## Goal

Automatically instrument navigation spans when using `react-native-navigation`

## Design

Add event listeners to the `Navigation` library to start and automatically end a navigation span when the route changes

**Note:** this approach also creates a navigation span for the initial route due to the way the end to end test is instrumented

## Changeset

- Add event listeners to plugin
- Add navigation span attributes

## Testing

- Update unit tests to check that a span in automatically started on route change
- Update end to end tests to check that a span is automatically started on a route change

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,13 +12,13 @@ steps:
   #
   # Upload each basic pipeline
   #
-  # - label: ":pipeline_upload: Basic browser pipeline"
-  #   commands:
-  #     - buildkite-agent pipeline upload .buildkite/browser-pipeline.yml
+  - label: ":pipeline_upload: Basic browser pipeline"
+    commands:
+      - buildkite-agent pipeline upload .buildkite/browser-pipeline.yml
 
-  # - label: ":pipeline_upload: React Native pipeline"
-  #   commands:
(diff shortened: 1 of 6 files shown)


gingerbenw:
@gareththackeray Should we know what the previous route was from a navigation?

gareththackeray:
> @gareththackeray Should we know what the previous route was from a navigation?

If we do know, I would suggest we include it as an attribute (`bugsnag.navigation.previous_route`).  If it is not possible or very difficult I think we can live without it.  We should also add it for React Navigation if we can.

yousif-bugsnag (approved):
LGTM - possibly out of scope for this PR but do we need an additional test to check that the app start instrumentation works?

gingerbenw:
> LGTM - possibly out of scope for this PR but do we need an additional test to check that the app start instrumentation works?

Added `PLAT-11721` to instrument and test app start spans

gingerbenw:
> > @gareththackeray Should we know what the previous route was from a navigation?
> 
> If we do know, I would suggest we include it as an attribute (`bugsnag.navigation.previous_route`). If it is not possible or very difficult I think we can live without it. We should also add it for React Navigation if we can.

Raised ticket `PLAT-11722` to add previous route attributes

PR #410, opened 2024-02-27
[PLAT-11567] Complete navigation component

gingerbenw:
## Goal

Add `CompleteNavigation` component to make navigation spans more accurate 

## Design

Same pattern for component lifecycle methods as the `@bugsnag/react-navigation` plugin component, but leveraging the `getPlugin` method from `BugsnagPerformance` as we are not able to provide a wrapper component and leverage context to have a single point of reference for number of mounted components and the state of the current navigation span

## Changeset

- Added new component
- Updated plugin implementation to end spans depending on the number of components waiting
- Added `bugsnag.navigation.ended_by` span attribute

## Testing

- Unit tests for the NavigationComplete component
- Unit test that the plugin handles awaiting components and updating span attributes correctly
- Update end to end test to instrument `NavigationComplete` component (there is currently a bug on iOS, so the assertion for the ended_by span attribute is currently loose and will be improved in a separate ticket as this affects both plugins)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,13 +12,13 @@ steps:
   #
   # Upload each basic pipeline
   #
-  # - label: ":pipeline_upload: Basic browser pipeline"
-  #   commands:
-  #     - buildkite-agent pipeline upload .buildkite/browser-pipeline.yml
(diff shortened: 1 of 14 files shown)


yousif-bugsnag (approved):
looks good!

PR #411, opened 2024-03-05
Add React Native 0.73 tests to CI

yousif-bugsnag:
## Goal

Adds React Native 0.73 tests to the CI test matrix

## Testing

Covered by CI


bin/generate-react-native-fixture
```diff
@@ -67,7 +67,7 @@ if (!process.env.SKIP_GENERATE_FIXTURE) {
   }
 
   // create the test fixture
-  const RNInitArgs = ['react-native', 'init', 'reactnative', '--package-name', 'com.bugsnag.fixtures.reactnative.performance', '--directory', fixtureDir, '--version', rnVersion, '--npm', '--skip-install']
+  const RNInitArgs = ['react-native@latest', 'init', 'reactnative', '--package-name', 'com.bugsnag.fixtures.reactnative.performance', '--directory', fixtureDir, '--version', rnVersion, '--npm', '--skip-install']
```
gingerbenw, line 70: does this need @latest? what version do we get without that?

gingerbenw (approved):
Looking good!

bin/generate-react-native-fixture
yousif-bugsnag: There was a bug in previous versions of the react native CLI where the component name was being set incorrectly in `MainActivity.kt` - this has only been fixed recently and just calling `npx react-native` wasn't pulling in the latest version (possibly due to npm caching?)

Would it be better to pin it to the most recent version instead?

gingerbenw: nah, that seems sensible enough - I was just surprised it didn't automatically use the latest! but as you say... caching.

PR #412, opened 2024-03-05
trace propagation

djskinner:
## Goal

Set `traceparent` header on intercepted network requests

## Design

- Use existing request-tracker interface into the fetch/xhr monkey patches

## Changeset

- Update request trackers to handle any `extraRequestHeaders` supplied by handlers
- Expand `NetworkRequestPlugin` to determine whether to set and value for `traceparent`
- `traceparent` request header will be added to all same-origin xhr/fetch requests by default
- `networkRequestCallback` behaviour changes:
  - returning `null` will neither create a span nor add traceparent headers
  - returning `{ url: null }` will disable creating a network span
  - returning `{ propagateTraceContext: boolean }` will override the default trace propagation behaviour.  

## Testing

- [x] unit test coverage
- [x] e2e test coverage
- [x] alpha release for manual testing against the dashboard


packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -38,12 +43,17 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
+  private trackRequest: RequestStartCallback = (startContext) => {
     if (!this.shouldTrackRequest(startContext)) return
 
     const networkRequestInfo = this.networkRequestCallback({ url: startContext.url, type: startContext.type })
 
-    if (!networkRequestInfo) return
+    if (!networkRequestInfo) {
+      return {
+        onRequestEnd: undefined,
+        extraRequestHeaders: this.getExtraRequestHeaders(startContext)
```
djskinner, line 54: where not tracking the request / creating a network span the request headers will be determined from the context

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -59,15 +69,49 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
+    return {
+      onRequestEnd: (endContext: RequestEndContext) => {
+        if (endContext.state === 'success') {
+          span.setAttribute('http.status_code', endContext.status)
+          this.spanFactory.endSpan(span, endContext.endTime)
+        }
+      },
+      extraRequestHeaders: this.getExtraRequestHeaders(
+        startContext,
+        span
```
djskinner, line 81: where a request is being tracked / network span created we use the supplied span

packages/platforms/browser/lib/config.ts
```diff
@@ -79,6 +80,11 @@ export function createSchema (hostname: string, defaultRoutingProvider: RoutingP
       defaultValue: defaultSendPageAttributes,
       message: 'should be an object',
       validate: isSendPageAttributes
+    },
+    tracePropagationUrls: {
+      defaultValue: [],
```
djskinner, line 85: Initially a default value of `[]` will be used so that it is disabled by default and we can publish this on the mainline - and enable in manually for dashboard and others as necessary. When we come to release proper, this will have a new defaultValue

> The default value of tracePropagationTargets is ['localhost', /^\//]. This means that by default, tracing headers are only attached to requests that contain localhost in their URL or requests whose URL starts with a '/' (for example GET /api/v1/users)

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -59,15 +69,49 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
   }
+
+  private getExtraRequestHeaders (startContext: RequestStartContext, span?: SpanInternal): Record<string, string> {
+    const extraRequestHeaders: Record<string, string> = {}
+
+    if (this.tracePropagationUrls.some(regexp => regexp.test(startContext.url))) {
+      if (span && span.parentSpanId) {
+        const traceId = span.traceId
+        const parentSpanId = span.parentSpanId
+        const sampled = span.samplingRate <= this.spanFactory.sampler.spanProbability.scaled
```
gareththackeray, line 97: Would this logic be better off in the sampler?

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -59,15 +69,49 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
     return startContext.url !== this.configEndpoint && permittedPrefixes.some((prefix) => startContext.url.startsWith(prefix))
   }
+
+  private getExtraRequestHeaders (startContext: RequestStartContext, span?: SpanInternal): Record<string, string> {
+    const extraRequestHeaders: Record<string, string> = {}
+
+    if (this.tracePropagationUrls.some(regexp => regexp.test(startContext.url))) {
+      if (span && span.parentSpanId) {
+        const traceId = span.traceId
+        const parentSpanId = span.parentSpanId
```
gareththackeray, line 96: It should be set to `span.id`.  It will be the _parent_ of the span created on the server.

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
```diff
@@ -59,15 +69,49 @@ export class NetworkRequestPlugin implements Plugin<BrowserConfiguration> {
...
 
-    return (endContext: RequestEndContext) => {
-      if (endContext.state === 'success') {
-        span.setAttribute('http.status_code', endContext.status)
-        this.spanFactory.endSpan(span, endContext.endTime)
-      }
+    return {
+      onRequestEnd: (endContext: RequestEndContext) => {
+        if (endContext.state === 'success') {
+          span.setAttribute('http.status_code', endContext.status)
```
gareththackeray, line 75: Couldn't `span` be null now?

djskinner: no

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
djskinner: which logic, this?

```
const sampled = currentSpanContext.samplingRate <= this.spanFactory.sampler.spanProbability.scaled
```

e.g. like this?

```
const sampled = this.spanFactory.sampler(currentSpanContext.samplingRate)
```

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
djskinner: Okay, I was wondering why the `traceparent` had the same value for both with and without a network span in the unit tests, this explains it. fixed in f7ded53

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
gareththackeray: Ah yes I see - the span null path in `getExtraRequestHeaders` is from the call on line 54.

packages/platforms/browser/lib/auto-instrumentation/network-request-plugin.ts
gareththackeray: I was more just thinking of a method on `sampler` called `sample` or `shouldSample`.  Just to have that logic (i.e. the `<=` test in a single place.

djskinner:
```
const sampled = this.spanFactory.sampler(currentSpanContext.samplingRate)
```

Should have been

```
const sampled = this.spanFactory.sampler.shouldSample(currentSpanContext.samplingRate)
```

So it would be used like this in the two places:

```
const sampled = this.spanFactory.sampler.shouldSample(currentSpanContext.samplingRate)
// or, if a span is being created
const sampled = this.spanFactory.sampler.shouldSample(span.samplingRate)
```

packages/platforms/browser/lib/config.ts
```diff
@@ -79,6 +80,11 @@ export function createSchema (hostname: string, defaultRoutingProvider: RoutingP
       defaultValue: defaultSendPageAttributes,
       message: 'should be an object',
       validate: isSendPageAttributes
+    },
+    tracePropagationUrls: {
```
tomlongridge, line 84: I feel like the JS SDK is a candidate for configuring through the callback, as is mentioned in the PD:

> We will decide on a SDK-by-SDK basis whether the config setting should be at the SDK level or at a network-request-instrumentation level.

Given that developers will be controlling whether to generate a span for a network request using the callback and `requestInfo` return value, does it feel odd to control the presence of the trace header by regex rather than a new field in `requestInfo` (e.g. `requestInfo.includeTraceHeader`?

gareththackeray: I agree with this. I think `propagateTraceContext` would be the right term.

gareththackeray (approved):

packages/request-tracker/lib/request-tracker.ts
```diff
@@ -31,16 +31,23 @@ export class RequestTracker {
...
     }
 
-    return (endContext: RequestEndContext) => {
-      for (const endCallback of endCallbacks) {
-        endCallback(endContext)
-      }
+    return {
+      onRequestEnd: (endContext: RequestEndContext) => {
+        for (const result of results) {
+          result?.onRequestEnd?.(endContext)
```
yousif-bugsnag, line 43: slight nit: I think we've been trying to avoid using optional chaining because it transpiles to a lot more code, e.g.
`(_a = result === null || result === void 0 ? void 0 : result.onRequestEnd) === null || _a === void 0 ? void 0 : _a.call(result, endContext)`

yousif-bugsnag (approved):
one minor nit, but otherwise LGTM 

not sure if the alpha release commit should be reverted though?

PR #415, opened 2024-03-07
[PLAT-11670] Navigation spans as children

gingerbenw:
## Goal

Allow Navigation spans to be children of other spans

This PR also separates versions of react native for testing into the 'full pipeline' to reduce build times with feature branches

## Testing

Update end to end tests to check navigation spans become children of the app start span
Update unit tests to ensure navigation spans can become children


.buildkite/react-native-navigation-pipeline.yml
```diff
@@ -56,7 +53,7 @@ steps:
 
       # iOS Build
       # ------------------------------
-      - label: ':building_construction: :mac: Build RN {{matrix.rn_version}} (Old Arch) test fixture'
```
yousif-bugsnag, line 59: Can we rename these to distinguish them from the regular react native tests?

.buildkite/react-native-pipeline.yml
```diff
@@ -17,7 +17,6 @@ steps:
         commands:
           - ./bin/generate-react-native-fixture
         matrix:
-          - "0.71"
```
yousif-bugsnag, line 20: may want to consider rebasing on next as 0.73 has been added to the pipeline now

yousif-bugsnag (approved):

PR #423, opened 2024-03-19
pass --dist-tag next to publish

djskinner:
## Goal

- `--dist-tag` is not a valid option for `version`, it needs to be passed to `publish`

## Changeset

- update the release script to pass `--dist-tag next` when publishing prereleases

## Testing

- performed a `2.3.0-alpha.0` of the `trace-propagation` branch using the updated script

### DIFF
--- a/bin/release.sh
+++ b/bin/release.sh
@@ -35,7 +35,7 @@ BROWSER_PACKAGE_CHANGED=$(npx lerna changed --parseable | grep -c packages/platf
 if [ -z "${RETRY_PUBLISH:-}" ]; then
   case $VERSION in
     "prerelease" | "prepatch" | "preminor" | "premajor")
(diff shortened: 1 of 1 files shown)


bin/release.sh
```diff
@@ -35,7 +35,7 @@ BROWSER_PACKAGE_CHANGED=$(npx lerna changed --parseable | grep -c packages/platf
 if [ -z "${RETRY_PUBLISH:-}" ]; then
   case $VERSION in
     "prerelease" | "prepatch" | "preminor" | "premajor")
-      npx lerna version "$VERSION"  --no-push --no-private --dist-tag next
+      npx lerna version "$VERSION"  --no-push --no-private
       ;;
```
gingerbenw, line 39: this is now a duplicate command to the other branch, so we can drop the branching logic here - otherwise all looks good!

gingerbenw (approved):

PR #426, opened 2024-03-19
[PLAT-11785] Update android namepsace

gingerbenw:
## Goal

To prevent an overlap with the native android library

## Changeset

Update the namepsace to `com.bugsnag.reactnative.performance`


lemnik (approved):
LGTM

PR #433, opened 2024-04-02
[PLAT-11800] Delay initial span batching

gingerbenw:
## Goal

This PR ensures that the correct sampling rate is used for a batch which is filled before the initial sampling request has been made by delaying batching while the request is in flight.

The config option for `samplingProbability` has also been removed.

## Testing

Add integration test for browser client which awaits the initial sampling request when a full batch is created before starting bugsnag

### DIFF
--- a/packages/core/lib/probability-manager.ts
+++ b/packages/core/lib/probability-manager.ts
@@ -14,7 +14,7 @@ class ProbabilityManager {
     const persistedProbability = await persistence.load('bugsnag-sampling-probability')
 
     let initialProbabilityTime: number
-    let initialTimoutDuration: number
+    let initialTimeoutDuration: number
 
     if (persistedProbability === undefined) {
       // If there is no stored probability:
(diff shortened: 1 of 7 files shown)


packages/core/lib/probability-manager.ts
```diff
@@ -102,6 +112,13 @@ class ProbabilityManager {
         if (lastProbabilityTimeBeforeTimeout === this.lastProbabilityTime) {
           this.setProbability(probability)
         }
+
+        // Initial sampling request has been made, and we can unblock batching
+        if (milliseconds === 0 && this.resolveInitialProbability) {
```
lemnik, line 117: The `milliseconds === 0` isn't doing much work, and could confuse things further down the road. I think it's probably safe to remove:

```suggestion
        if (this.resolveInitialProbability) {
```

lemnik (approved):
LGTM - With a trivial suggestion

PR #434, opened 2024-04-07
Update request-tracker-fetch.ts

tpenaranda:
fetch() override is killing RTK Query headers. As soon I call BugsnagPerformance.start() all my RTK custom headers are lost. ('Content-Type', 'Accept'). Then, it completely breaks the project since the BE replies with http 415. 

## Goal

Let BugsnagPerformance to play well with a project setting custom headers at fetch() call input. 

## Design

Looks like that's the way how we set headers on a fetch() request. https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options

## Changeset

So far I see how RequestTracker() is constructed, 'extraRequestHeaders' variable is always an empty array, and those headers set by the original fetch() call are set on 'input' variable (function param), then I'm suggesting to send 'input.headers' value to 'mergeRequestHeaders' call. Finally, those headers are merged in with the ones at 'init'.
Honestly, I'm not 100% sure this is the right fix, but I'm pretty sure it's gonna help to understand/find/solve the issue. Additionally, IMHO, this global.fetch thing is a bit reckless.

## Testing

Manual testing. Before fix, no headers are sent... later, everything is there.

### DIFF
--- a/packages/request-tracker/lib/request-tracker-fetch.ts
+++ b/packages/request-tracker/lib/request-tracker-fetch.ts
@@ -25,10 +25,10 @@ function createFetchRequestTracker (global: GlobalWithFetch, clock: Clock) {
   global.fetch = function fetch (input?: unknown, init?: unknown) {
     const startContext = createStartContext(clock.now(), input, init, global.document && global.document.baseURI)
 
-    const { onRequestEnd, extraRequestHeaders } = requestTracker.start(startContext)
+    const { onRequestEnd } = requestTracker.start(startContext)
 
     // Add the headers to the `init` received from the caller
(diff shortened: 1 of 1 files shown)


clr182:
Hi @tpenaranda 

Thank you for reaching out and raising this bug report. 

This does seem to be an issue on our side that we will look to address soon. Once we have more information we will be sure to reach out and update you via this ticket thread. 

Thanks!

yousif-bugsnag:
Thanks again for raising this @tpenaranda - I'm going to close this in favour of #436, which we'll look to get out in the next release.

PR #438, opened 2024-04-17
[PLAT-12014] Update @bugsnag/cuid

gingerbenw:
## Goal

Use validation and type definitions from `@bugsnag/cuid` v3.1.0


yousif-bugsnag (approved):

gingerbenw:
@yousif-bugsnag do you think this needs a changelog entry? it doesn't really affect much of anything so I'm thinking it's not particularly useful

PR #445, opened 2024-04-25
fix cuid imports

djskinner:
## Goal

Fixes the following error when using esm in node: Named export 'isCuid' not found. The requested module '@bugsnag/cuid' is a CommonJS module, which may not support all module.exports as named exports. CommonJS modules can always be imported via the default export

## Testing

- Reproduced the error using a "qwik" project as reported by one of our customers. Then made the same changes manually to the files in node_modules to verify the error is resolved


gingerbenw (approved):
is this considered a bugfix? could probably do with a changelog entry

PR #446, opened 2024-04-26
set parent context of full page loads from traceparent meta

djskinner:
## Goal

> Aligning with the [existing OpenTelemetry advice](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/7c7294ce0591a3c8d4c12b5f135f4fcd24b79762/plugins/web/opentelemetry-instrumentation-document-load/README.md#optional-send-a-trace-parent-from-your-server) for this response propagation, we will use a meta tag in the page HTML, with name="traceparent" and content as the standard traceparent string to extract traceparent and set the traceid and root span parentid accordingly.

## Design

- in the full page load plugin: If the traceparent meta tag is present in the HTML document, parse it and use it as the `parentContext` when creating the full page load span

## Testing

- unit tests
- e2e tests


packages/platforms/browser/lib/auto-instrumentation/full-page-load-plugin.ts
```diff
@@ -54,7 +55,21 @@ export class FullPageLoadPlugin implements Plugin<BrowserConfiguration> {
       return
     }
 
-    const span = this.spanFactory.startSpan('[FullPageLoad]', { startTime: 0, parentContext: null })
+    let parentContext: ParentContext | null = null
+
+    const traceparentMetaTag = document.querySelector('meta[name="traceparent"]')
+    if (traceparentMetaTag !== null && traceparentMetaTag.getAttribute('content')) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const traceparent = traceparentMetaTag.getAttribute('content')!
```
gingerbenw, line 63: is this non-null assertion necessary, based on the logic above?

djskinner: yes, as there is nothing to stop a function returning something different on a subsequent call, so TypeScript makes no assumptions. Maybe there's a nicer way to write this? I recall thinking about doing an assignment in the if statement but wasn't keen on that personally

djskinner: Well, its much nicer with optional chaining but last time I used that I was asked to reverse it, I think due to impact on bundle size

gingerbenw: okay, that's totally fine then - I was just surprised this could be null at this stage based on the if statement.

gingerbenw (approved):
do we need a changelog entry for this one?

PR #448, opened 2024-04-30
[PLAT-11983] Manual network span API

gingerbenw:
## Goal

add a new `startNetworkSpan` method to the core client
update network span naming convention to `[HTTP/VERB]`

## Testing

Add unit test for `startNetworkSpan` and attributes

### DIFF
--- a/packages/core/tests/core.test.ts
+++ b/packages/core/tests/core.test.ts
@@ -314,7 +314,7 @@ describe('Core', () => {
           await jest.runOnlyPendingTimersAsync()
 
           expect(delivery).toHaveSentSpan(expect.objectContaining({
-            name: '[HTTP]GET',
+            name: '[HTTP]/GET',
             kind: 3,
             events: [],
             spanId: 'a random 64 bit string',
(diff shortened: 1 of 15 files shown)


packages/core/tests/core.test.ts
```diff
@@ -295,6 +296,40 @@ describe('Core', () => {
...
+          await jest.runOnlyPendingTimersAsync()
+
+          expect(delivery).toHaveSentSpan(expect.objectContaining({
+            name: '[HTTP]/GET',
+            kind: 3,
+            events: [],
+            spanId: 'a random 64 bit string',
+            traceId: 'a random 128 bit string',
+            startTimeUnixNano: expect.any(String),
+            endTimeUnixNano: expect.any(String)
```
tomlongridge, line 323: I think we'd expect page and referrer attributes on spans from browsers - do you think that's reasonably covered elsewhere to validate that it works for these manual spans or can we add an additional assertion for it?

gingerbenw: we have this covered already in the span-attributes-source browser tests and network-spans end to end tests for browser

tomlongridge (approved):

PR #461, opened 2024-06-14
[PLAT-12231] Browser error correlation

gingerbenw:
## Goal

Implement error correlation for browser performance

## Design

Call `addOnError` from bugsnag notifier when provided as a configuration option

## Changeset

Added new `bugsnag` configuration option to core client

## Testing

Added unit tests for adding span and trace id to a reported error

### DIFF
--- a/packages/core/lib/config.ts
+++ b/packages/core/lib/config.ts
@@ -1,6 +1,16 @@
 import { type Plugin } from './plugin'
 import { isLogger, isNumber, isObject, isPluginArray, isString, isStringArray, isStringWithLength } from './validation'
 
+interface BugsnagErrorEvent {
+  addMetadata: (tabName: string, data: Record<string, any>) => void
+}
+
+type BugsnagErrorCallback = (event: BugsnagErrorEvent) => void
+
+interface BugsnagErrorStatic {
+  addOnError: (fn: BugsnagErrorCallback) => void
+}
+
 export interface Logger {
   debug: (message: string) => void
   info: (message: string) => void
@@ -16,6 +26,7 @@ export interface Configuration {
   appVersion?: string
   enabledReleaseStages?: string[] | null
   plugins?: Array<Plugin<Configuration>>
+  bugsnag?: BugsnagErrorStatic
 }
 
 export interface TestConfiguration {
@@ -42,6 +53,7 @@ export interface CoreSchema extends Schema {
   appVersion: ConfigOption<string>
   enabledReleaseStages: ConfigOption<string[] | null>
   plugins: ConfigOption<Array<Plugin<Configuration>>>
+  bugsnag: ConfigOption<BugsnagErrorStatic | undefined>
 }
 
 export const schema: CoreSchema = {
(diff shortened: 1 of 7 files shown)


packages/core/lib/config.ts
```diff
@@ -1,6 +1,16 @@
 import { type Plugin } from './plugin'
 import { isLogger, isNumber, isObject, isPluginArray, isString, isStringArray, isStringWithLength } from './validation'
 
+interface BugsnagErrorEvent {
+  setTraceCorrelation: (traceId: string, spanId: string) => void
```
imjoehaines, line 5: I'd suggest marking this optional as it could be missing

```suggestion
  setTraceCorrelation?: (traceId: string, spanId: string) => void
```

packages/plugin-react-navigation/lib/create-navigation-container.tsx
```diff
@@ -27,7 +27,7 @@ export const createNavigationContainer: CreateNavigationContainer = (NavigationC
       }
 
       if (typeof onStateChange === 'function') {
-        onStateChange.apply(this, args)
+        onStateChange.apply(undefined, args)
```
imjoehaines, line 30: Is this change related?

imjoehaines (commented):
Could we add a maze runner test for this?

packages/plugin-react-navigation/lib/create-navigation-container.tsx
gingerbenw: no, actually! I can separate this out

gingerbenw:
> Could we add a maze runner test for this?

Covered in PLAT-12233

packages/core/lib/core.ts
```diff
@@ -68,6 +68,16 @@ export function createClient<S extends CoreSchema, C extends Configuration, T> (
     start: (config: C | string) => {
       const configuration = validateConfig<S, C>(config, options.schema)
 
+      if (configuration.bugsnag) {
```
imjoehaines, line 71: Is it worth checking here that `setTraceCorrelation` exists so we avoid even adding the callback if the bugsnag-js version is too old?

imjoehaines (approved):

packages/core/lib/core.ts
gingerbenw: oooh, that's even better!

gingerbenw: although, it appears I'm not entirely sure how to check for that...

imjoehaines: I think you can check for `configuration.bugsnag.Event.prototype.setTraceCorrelation`:

https://github.com/bugsnag/bugsnag-js/blob/76f50d9b0684d8a445829c932fc182ee191775ab/packages/browser/src/notifier.js#L105

gingerbenw: Do you think that would be best placed here, or in the config validation?

imjoehaines: Sorry, missed this!

I think here is best — if/when we add more stuff that needs APIs from `Bugsnag`, we probably don't want to disable _all_ the error integrations if new APIs are missing

gingerbenw: Raised PLAT-12350 to cover this, so we can get this ticket merged down

PR #462, opened 2024-06-18
[PLAT-12233] Error correlation browser testing

gingerbenw:
## Goal

Add end to end tests for error correlation in browser

### DIFF
--- a/test/browser/features/error-correlation.feature
+++ b/test/browser/features/error-correlation.feature
@@ -1,0 +1,23 @@
+Feature: Error Correlation
+
+  Scenario: Reported errors include the current trace and span id
+    Given I navigate to the test URL "/error-correlation"
+    And I wait to receive a sampling request
+    Then I click the element "start-span"
+    And I click the element "send-error"
+
+    # Wait for the error
+    And I wait to receive an error
+    Then the error is a valid browser payload for the error reporting API
+
+    # End the span
+    Then I click the element "end-span"
+    And I wait for 1 span
+
+    # Assert on the span, and store the trace id and span id
+    Then the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.id" is stored as the value "spanId"
+    And the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.traceid" is stored as the value "traceId"
+
+    # Check span id is included in the error
+    And the error payload field "events.0.correlation.spanId" equals the stored value "spanId"
+    And the error payload field "events.0.correlation.traceId" equals the stored value "traceId"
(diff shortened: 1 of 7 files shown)


test/browser/features/fixtures/packages/error-correlation/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "error-correlation",
+    "private": true,
+    "scripts": {
+        "build": "rollup --config ../rollup.config.mjs",
+        "clean": "rm -rf dist"
+    },
+    "dependencies": {
+        "@bugsnag/browser": "^7.25.0-alpha.0"
```
yousif-bugsnag, line 9: is there a particular reason for using the `@bugsnag/browser` dependency directly here and not `@bugsnag/js`?

gingerbenw: only that it's in the `browser` fixture, and not node, meaning for a smaller install

yousif-bugsnag: ah okay, makes sense 👍🏾

test/browser/features/error-correlation.feature
```diff
@@ -0,0 +1,51 @@
...
+
+    # Assert on the span, and store the trace id and span id
+    Then the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.id" is stored as the value "spanId"
+    And the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.traceid" is stored as the value "traceId"
+
+    # Check span id is included in the error
+    And the error payload field "events.0.correlation.spanId" equals the stored value "spanId"
+    And the error payload field "events.0.correlation.traceId" equals the stored value "traceId"
+
+  Scenario: Reported errors include the parent trace and span id
```
yousif-bugsnag, line 25: I found this scenario a bit confusing as it reads like the error should always include the parent trace and span id when there's a nested span.

Maybe something like 'Reported errors do not include the trace and span id if the span is not the current context' - or maybe even roll them both into one scenario?

Also might be worth adding a test for when there are no open spans?

test/browser/features/error-correlation.feature
```diff
@@ -0,0 +1,63 @@
...
+
+    # Assert on the span, and store the trace id and span id
+    Then the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.id" is stored as the value "spanId"
+    And the trace payload field "resourceSpans.0.scopeSpans.0.spans.0.traceid" is stored as the value "traceId"
+
+    # Check span id is included in the error
+    And the error payload field "events.0.correlation.spanId" equals the stored value "spanId"
+    And the error payload field "events.0.correlation.traceId" equals the stored value "traceId"
+
+  Scenario: Reported errors do not include the trace and span id if the span is not the current context
```
gingerbenw, line 33: @yousif-bugsnag  does this make a bit more sense? and am I missing something simple here? it's failing on like 62 at the moment

yousif-bugsnag: yeah that definitely makes more sense 👍🏾 

not sure why that is failing though as it seems sensible enough 🤔 would need to take a look at the traces to debug it

lemnik (approved):
LGTM

PR #465, opened 2024-06-21
Move 'full' React Native pipeline to MacOS 14 / Java 17

yousif-bugsnag:
## Goal

Move the 'full' react native CI pipeline to MacOS 14/Java 17 (required for React Native 0.73) - this was missed in #464 when 0.73 tests were moved from the basic to full pipeline.

## Testing

covered by a full CI run

### DIFF
--- a/.buildkite/react-native-pipeline.full.yml
+++ b/.buildkite/react-native-pipeline.full.yml
@@ -8,8 +8,9 @@ steps:
         key: "build-react-native-android-fixture-old-arch-full"
         timeout_in_minutes: 30
         agents:
-          queue: macos-12-arm
+          queue: macos-14
         env:
+          JAVA_VERSION: "17"
           NODE_VERSION: "18"
           RN_VERSION: "{{matrix}}"
           BUILD_ANDROID: "true"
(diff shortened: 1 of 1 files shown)


.buildkite/react-native-pipeline.full.yml
```diff
@@ -53,12 +55,12 @@ steps:
...
         timeout_in_minutes: 30
         agents:
-          queue: "macos-12-arm"
+          queue: "macos-14"
         env:
           NODE_VERSION: "18"
           RN_VERSION: "{{matrix}}"
           BUILD_IOS: "true"
-          DEVELOPER_DIR: "/Applications/Xcode14.app"
+          XCODE_VERSION: "15.3.0"
```
gingerbenw, line 63: This no longer needs the full path?

gingerbenw (approved):

.buildkite/react-native-pipeline.full.yml
yousif-bugsnag: nope! the new build servers have moved over to using `XCODE_VERSION` instead of `DEVELOPER_DIR` - which is handy because you don't need to know where it's installed

gingerbenw: sublime! 👌🏻

PR #469, opened 2024-07-01
Fix crash when navigation ref is created with useRef

yousif-bugsnag:
## Goal

Support navigation refs created using `useRef`.

Currently the React Navigation plugin expects the passed in navigation ref to be created with `useNavigationContainerRef`, and crashes if the ref was created with `useRef`, since in that case `getCurrentRoute` should be accessed via `ref.current`.

## Design

the code now first checks if `ref.current` exists and then falls back to calling `getCurrentRoute` on the ref directly.

### DIFF
--- a/packages/plugin-react-navigation/lib/create-navigation-container.tsx
+++ b/packages/plugin-react-navigation/lib/create-navigation-container.tsx
@@ -19,9 +19,9 @@ export const createNavigationContainer: CreateNavigationContainer = (NavigationC
     const routeNameRef = useRef<string>()
 
     const wrappedOnStateChange: typeof onStateChange = (...args) => {
-      const currentRoute = navigationContainerRef && navigationContainerRef.current
+      const currentRoute = navigationContainerRef ? navigationContainerRef.current
         ? navigationContainerRef.current.getCurrentRoute()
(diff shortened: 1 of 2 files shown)


packages/plugin-react-navigation/lib/create-navigation-container.tsx
```diff
@@ -19,7 +19,9 @@ export const createNavigationContainer: CreateNavigationContainer = (NavigationC
     const routeNameRef = useRef<string>()
 
     const wrappedOnStateChange: typeof onStateChange = (...args) => {
-      const currentRoute = navigationContainerRef ? navigationContainerRef.getCurrentRoute() : null
+      const currentRoute = navigationContainerRef && navigationContainerRef.current
+        ? navigationContainerRef.current.getCurrentRoute()
+        : navigationContainerRef ? navigationContainerRef.getCurrentRoute() : null
```
gingerbenw, line 24:
```suggestion
      const currentRoute = navigationContainerRef ? navigationContainerRef.current
        ? navigationContainerRef.current.getCurrentRoute()
        : navigationContainerRef.getCurrentRoute() : null
```

gingerbenw (approved):

PR #471, opened 2024-07-12
[PLAT-12418] TypeScript definition 3.8 support

gingerbenw:
## Goal

Ensure TS definitions are compatible with earlier versions of TypeScript (down to 3.8) by enforcing top level type only imports

*Note*: There is an outstanding TODO left in the eslint config, as I have been unable to find a plugin that works in the same way for type only exports. If a new export is added and determined to be type only, the config will automatically add a top level `export types` pattern - however if an inline one is manually added, the current configuration will not correct it.

## Testing

Built and installed in a local app using TypeScript 3.8 - project would not build until changes were implemented due to inline type imports, such as `import { type MyType } from 'module'`


.eslintrc.js
```diff
@@ -31,7 +31,11 @@ const tsRuleOverrides = {
   '@typescript-eslint/no-for-in-array': 'off',
 
   // Optional chaining compiles to a lot more code
-  '@typescript-eslint/prefer-optional-chain': 'off'
+  '@typescript-eslint/prefer-optional-chain': 'off',
+
+  // Support TypeScript 3.8 by disallowing import { type Module } from 'module'
+  // TODO: Find the export equivalent
```
djskinner, line 37: how about https://typescript-eslint.io/rules/consistent-type-exports/ ?

djskinner (approved):

.eslintrc.js
gingerbenw: I did try this one! It gives the same behaviour as we have now, where it will allow `export { type ReadonlySampler, type ReadWriteSampler }` if you type it manually

PR #474, opened 2024-07-23
[PLAT-12475] Monkey patch notify method 

gingerbenw:
## Goal

Refactor the error correlation implementation by patching the `_notify` method instead of leveraging `addOnError` to ensure callbacks run before the current span has closed


yousif-bugsnag (approved):
LGTM

Issue #460, opened 2024-06-12
Vue router path dynamic param with empty parenthesis doesn't match corresponding url


thomaspaillot:
### Describe the bug
Vue router route path can contain parenthesis at the end of a dynamic param like this /:projects()/all and when you use path-to-regexp to match a potential url the regex generated by path-to-regexp doesn't work because of the parenthesis.

### Steps to reproduce
```
import pathToRegexp from 'path-to-regexp';

'/my-project-name/all'.match(pathToRegexp('/:projects()/all')) // return false
'/my-project-name/all'.match(pathToRegexp('/:projects(\\S+)/all')) // return true
'/my-project-name/all'.match(pathToRegexp('/:projects/all')) // return true
````

### Environment
* BugSnag Performance library version: 2.6.0
* BugSnag Vue Router Performance library version: 2.4.0
* Browser framework version (if any):
    * Vue: 3.4.27

### Example Repo

[...]

PR #476, opened 2024-07-24
feat(vue-router): use vue router to resolve routes

yousif-bugsnag:
## Goal

Replace the usage of `path-to-regexp` and use Vue's `router.resolve` to resolve routes in `VueRoutingProvider`.

Fixes an issue where routes with parantheses (generated by nuxt) were failing to resolve (see #460)

## Design

v4 of `vue-router` (which we depend on) no longer uses `path-to-regexp` internally to parse routes, and now uses its own parsing system that supports dynamic routing. 

The router object has a `resolve` method that seems to do what we need (find a corresponding route from a given URL) - and also removes the need to flatten routes when resolving.

## Changeset

- Removed path-to-regexp dependency
- Updated resolveRoute to use Vue's router.resolve
- Updated e2e tests to test nested routes and routes with parentheses

## Testing

Updated the e2e tests to include a nested route and a route with parentheses

### DIFF
--- a/packages/vue-router/lib/vue-router-routing-provider.ts
+++ b/packages/vue-router/lib/vue-router-routing-provider.ts
@@ -1,21 +1,6 @@
-import {
-  onSettle
-
-} from '@bugsnag/browser-performance'
+import { onSettle } from '@bugsnag/browser-performance'
 import type { RouteResolver, RoutingProvider, StartRouteChangeCallback } from '@bugsnag/browser-performance'
-import pathToRegexp from 'path-to-regexp'
-import type { RouteRecordRaw, Router } from 'vue-router'
-
-function flattenRoutes (routes: RouteRecordRaw[], _prefix: string = ''): string[] {
-  const prefix = `${!_prefix || _prefix === '/' ? _prefix : `${_prefix}/`}`
-  return [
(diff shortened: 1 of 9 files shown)


packages/vue-router/lib/vue-router-routing-provider.ts
```diff
@@ -27,7 +12,8 @@ export class VueRouterRoutingProvider implements RoutingProvider {
     const normalizedBasename = !basename || basename === '/' ? '' : basename
 
     function resolveRoute (url: URL): string {
-      return flattenRoutes(router.getRoutes()).find((fullRoutePath) => url.pathname.replace(normalizedBasename ?? '', '').match(pathToRegexp(fullRoutePath))) || 'no-route-found'
+      const location = router.resolve({ path: url.pathname.replace(normalizedBasename ?? '', '') })
+      return location.matched.length > 0 ? location.matched[location.matched.length - 1].path : 'no-route-found'
```
imjoehaines, line 16: nit:

```suggestion
      return location.matched.pop() || 'no-route-found'
```

imjoehaines (approved):

gingerbenw (approved):

gingerbenw:
Don't forget the changelog entry! 👍🏻

PR #478, opened 2024-07-26
React Native: Ensure native module is fully backwards compatible

yousif-bugsnag:
## Goal

Refactors the native module in the React Native package to ensure full backwards compatibility with the Old Architecture.

## Design

- Android codebase has been updated so that the New Architecture types are not compiled when not available. 
- iOS podspec has been updated to ensure the correct dependencies are installed on Old Architecture.
- `native.ts` has been updated to load the native module from the old NativeModules API if turbo modules are not enabled. 

## Testing

Covered by a full CI run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,13 @@
 
 ### Added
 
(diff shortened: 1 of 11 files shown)


gingerbenw (commented):
Looks good from my side - but would be keen to have @lemnik review as well 👍🏻

lemnik (approved):

PR #481, opened 2024-07-30
Avoid long running timers when ensuring probability freshness

imjoehaines:
## Goal

Currently we set long running timers when the `ProbabilityManager` is created or when a probability value is set in order to ensure we check for a fresh probability value at least once a day. This is necessary because if the probability is set to 0 by the server then this is the only mechanism that will eventually reset the probability to a non-zero value

The current design leads to a warning on some versions of React Native because there are possible correctness concerns — the timers aren't triggered in the background and so won't fire until the app returns to the foreground. For our purposes this is fine, but we can avoid the warning with a small refactor

## Design

This PR replaces the current `setTimeout` mechanism with a new `ProbabilityManager#ensureFreshProbability` method. This method can be called by consuming classes, e.g. the `BatchProcessor`, to ensure that the probability value is less than 24 hours old. If it is stale then a new probability is fetched from the server

This is also called by the backgrounding listener when the app returns to the foreground, to ensure a non-zero probability value will eventually be fetched when the probability has been set to 0

### DIFF
--- a/packages/core/lib/batch-processor.ts
+++ b/packages/core/lib/batch-processor.ts
@@ -6,7 +6,7 @@ import type { RetryQueue } from './retry-queue'
 import type { ReadonlySampler } from './sampler'
 import type { SpanEnded } from './span'
 
-type MinimalProbabilityManager = Pick<ProbabilityManager, 'setProbability'>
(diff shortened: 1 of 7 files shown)


yousif-bugsnag (approved):

gingerbenw (approved):
Fresh! 

Neat refactor 👍🏻

PR #486, opened 2024-08-05
refactor(react-native): use batch time for filenames in retry queue

yousif-bugsnag:
## Goal

Replace usage of BigInt in the react native retry queue as this isn't supported on older versions of React Native (or more specifically on older versions of Hermes).

## Design

Instead of using nanosecond timestamps based on the largest span end time in the payload, the retry queue now uses the time the payload was added to the queue (in ms) for the filename. This is already being passed to `FileBasedRetryQueue#add` by the batch processor, but was previously an unused parameter.

Although it shouldn't really be possible for two files to be created with the same timestamp (as the time is recorded when a batch is flushed and a failed batch is only ever added to the retry queue once), the span ID of the last span in the payload is still used in the filename to ensure filenames are unique.

This also ensures backwards compatibility - the `timestampFromFilename` function used to parse filenames now simply returns a number instead of a bigint, so any existing retry files using nanosecond timestamps will still be parsed.

## Testing

Updated unit tests and relied on existing e2e tests

### DIFF
--- a/packages/platforms/react-native/lib/retry-queue/file-based.ts
+++ b/packages/platforms/react-native/lib/retry-queue/file-based.ts
@@ -8,25 +8,18 @@ import type {
 import type Directory from './directory'
 import timestampFromFilename from './timestamp-from-filename'
 
-function getLatestSpan (body: DeliveryPayload): DeliverySpan | undefined {
-  let latestSpan: DeliverySpan | undefined
-  let biggestTimestamp: bigint | undefined
+function getLastSpan (body: DeliveryPayload): DeliverySpan | undefined {
+  let lastSpan: DeliverySpan | undefined
 
   for (const resourceSpan of body.resourceSpans) {
     for (const scopeSpan of resourceSpan.scopeSpans) {
       for (const span of scopeSpan.spans) {
-        if (biggestTimestamp === undefined || BigInt(span.endTimeUnixNano) > biggestTimestamp) {
-          latestSpan = span
-
-          // store the biggest timestamp separately as a bigint so we don't need
-          // to parse it on every iteration
-          biggestTimestamp = BigInt(latestSpan.endTimeUnixNano)
-        }
+        lastSpan = span
       }
     }
   }
 
-  return latestSpan
+  return lastSpan
 }
 
 function isValidFilename (filename: string): boolean {
@@ -42,18 +35,18 @@ function isValidFilename (filename: string): boolean {
     return false
   }
 
-  const nowInNanoseconds = BigInt(Date.now()) * NANOSECONDS_IN_MILLISECONDS
+  const now = Date.now()
 
   // files that are older than 24 hours are not valid as the data may not be
   // relevant anymore
-  if (timestamp < nowInNanoseconds - NANOSECONDS_IN_DAY) {
+  if (timestamp < now - MILLISECONDS_IN_DAY) {
     return false
   }
 
   // files that have a timestamp more than 24 hours in the future are not
   // valid as something may have gone wrong with the clock and this is
   // unlikely to be a timezone issue
-  if (timestamp > nowInNanoseconds + NANOSECONDS_IN_DAY) {
+  if (timestamp > now + MILLISECONDS_IN_DAY) {
     return false
   }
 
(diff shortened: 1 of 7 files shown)


packages/platforms/react-native/lib/retry-queue/file-based.ts
```diff
@@ -8,25 +8,18 @@ import type {
...
-          latestSpan = span
-
-          // store the biggest timestamp separately as a bigint so we don't need
-          // to parse it on every iteration
-          biggestTimestamp = BigInt(latestSpan.endTimeUnixNano)
-        }
+        lastSpan = span
       }
     }
   }
```
gingerbenw, line 20: now we're not searching for the biggest timestamp, wouldn't this be more performance to get the very last span by index, or using a similar array method?

packages/platforms/react-native/tests/retry-queue/directory.test.ts
```diff
@@ -11,15 +11,17 @@ describe('RetryQueueDirectory', () => {
...
       ])
 
       expect(await directory.files()).toStrictEqual([
         'retry-4-z.json',
-        'retry-3-a.json',
+        'retry-3-b.json',
         'retry-2-x.json',
+        'retry-1-a.json',
         'retry-1-d.json'
       ])
```
gingerbenw, line 26: I might be missing something - why did this change?

packages/platforms/react-native/tests/retry-queue/timestamp-from-filename.test.ts
```diff
@@ -9,12 +9,9 @@ describe('timestampFromFilename', () => {
...
     ['retry-999999999999999999999999999999999999-a.json', '999999999999999999999999999999999999']
   ])('parses %s into the timestamp %s', (input, expected) => {
-    // ideally we'd want to do something like this:
-    // expect(timestampFromFilename(input)).toBe(BigInt(expected))
-    // but Jest can't serialise BigInts yet so we have to use strings for now
     const actual = timestampFromFilename(input)
-
-    expect(actual?.toString()).toBe(expected)
+    expect(actual).toBe(Number(expected))
+    expect(Number.isNaN(actual)).toBe(false)
```
gingerbenw, line 14:
```suggestion
    expect(typeof actual).toBe('number')
```

?

packages/platforms/react-native/tests/retry-queue/directory.test.ts
yousif-bugsnag: Just to make sure the filenames are being sorted as expected if the timestamps are equal

packages/platforms/react-native/tests/retry-queue/timestamp-from-filename.test.ts
```diff
@@ -9,12 +9,9 @@ describe('timestampFromFilename', () => {
     ['retry-00000000000-a.json', '0'],
     ['retry-999999999999999999999999999999999999-a.json', '999999999999999999999999999999999999']
   ])('parses %s into the timestamp %s', (input, expected) => {
-    // ideally we'd want to do something like this:
-    // expect(timestampFromFilename(input)).toBe(BigInt(expected))
-    // but Jest can't serialise BigInts yet so we have to use strings for now
     const actual = timestampFromFilename(input)
-
-    expect(actual?.toString()).toBe(expected)
+    expect(actual).toBe(Number(expected))
```
imjoehaines, line 13: nit: we could change `expected` to a number so that we don't need to parse it as one here

packages/platforms/react-native/tests/retry-queue/timestamp-from-filename.test.ts
imjoehaines:
```suggestion
    expect(actual).not.toBeNaN()
```

https://jestjs.io/docs/expect#tobenan

gingerbenw: > `typeof NaN` is `number`!

🤦🏻 thanks, JavaScript.

packages/platforms/react-native/tests/retry-queue/timestamp-from-filename.test.ts
yousif-bugsnag: yeah good point - that also removes the need for the `NaN` check

packages/platforms/react-native/lib/retry-queue/file-based.ts
yousif-bugsnag: Good point - updated to loop backwards and get the last span by index

packages/platforms/react-native/tests/retry-queue/timestamp-from-filename.test.ts
yousif-bugsnag: I've made the expected values numbers so this check is redundant

imjoehaines (approved):

PR #487, opened 2024-08-06
add a fixed sampling probability configuration option

djskinner:
## Goal

To support non-BugSnag trace endpoints, such as an Otel collector with tail-based sampling, we should provide the client RUM SDKs with a samplingProbability configuration that can be set by the developer to provide a fixed p-value for sampling.

## Design

Add a `samplingProbability` configuration option, which, when present, disables the bugsnag persisted probability fetching logic and always uses the supplied value for sampling probability.

## Changeset

- Add a `samplingProbability` configuration option
- Add FixedProbabilityManager
- Use FixedProbabilityManager when `samplingProbability` is defined

## Testing

- [x] unit tests
- [ ] e2e tests
- [ ] manual test against dashboard


packages/core/lib/config.ts
```diff
@@ -111,6 +113,11 @@ export const schema: CoreSchema = {
     defaultValue: undefined,
     message: 'should be an instance of Bugsnag',
     validate: (value: unknown): value is BugsnagErrorStatic => isObject(value) && typeof value.addOnError === 'function'
+  },
+  samplingProbability: {
+    defaultValue: undefined,
+    message: 'should be a number',
+    validate: (value: unknown): value is number | undefined => value === undefined || isNumber(value)
```
yousif-bugsnag, line 120: should this be a number between 0 and 1?

yousif-bugsnag (approved):
One query about the config validation but otherwise LGTM

PR #491, opened 2024-08-16
Set initial background state to foreground if initial app state is 'inactive'

yousif-bugsnag:
## Goal

In React Native 0.74+, when an iOS app launches the initial `AppStateStatus` changed from `unknown` to `inactive`. This wasn't being accounted for in the backgrounding listener, so the initial background state was incorrectly being set to `in-background`. 

This PR updates the React Native backgrounding listener so that the app is considered in foreground on launch unless the initial app state is `background`.

## Design

When the backgrounding listener is created the initial app state is now only set to `in-background` if the initial app state status is explicitly `background` - any other initial state (`active`, `inactive`, or `unknown`) is considered `in-foreground`  until an app state change event occurs.

## Testing

Covered by unit tests and manual testing.

The current structure of the end-to-end test fixture means that we can't test this effectively since the app state status is always 'active' when the scenario runs. This will be addressed in a later PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - Update error correlation implementation to monkey patch the error notifier [#474](https://github.com/bugsnag/bugsnag-js-performance/pull/474)
 - (react-native) Ensure native module is fully backwards compatible [#478](https://github.com/bugsnag/bugsnag-js-performance/pull/478)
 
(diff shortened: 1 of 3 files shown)


gingerbenw (approved):
Looks good, but needs a changelog entry 👍🏻

PR #493, opened 2024-08-27
instrument network requests with tracestate header

djskinner:
## Goal

Add `tracestate` header alongside existing `traceparent` for distributed tracing

## Design

Uses the same approach as `traceparent`

## Changeset

- network request plugin now includes an additional `tracestate` header alongside existing `traceparent` when configured to do so

## Testing

- unit tests
- e2e tests


yousif-bugsnag (approved):
LGTM

yousif-bugsnag:
Just realised that this is implemented for browser only - is that by design or should it be implemented for React Native also?

djskinner:
> Just realised that this is implemented for browser only - is that by design or should it be implemented for React Native also?

As discussed, we're doing this for browser only for now

test/browser/features/trace-propagation.feature
```diff
@@ -1,78 +1,84 @@
...
-        Given I navigate to the test URL "/traceparent"
+    Scenario: traceparent and tracestate header is added to xhr requests
+        Given I navigate to the test URL "/trace-propagation"
         When I click the element "xhr"
 
         And I wait to receive a reflection
         Then the reflection request method equals "GET"
         And the reflection "X-Test-Header" header equals "test"
         And the reflection "traceparent" header matches the regex "^00-[A-Fa-f0-9]{32}-[A-Fa-f0-9]{16}-01"
+        And the reflection "tracestate" header matches the regex "^sb=v:1;r:[0-9]{0,32}"
```
imjoehaines, line 11: I don't think this should ever be 0 characters — wouldn't it always be at least 1?

imjoehaines (approved):

PR #494, opened 2024-08-28
Discard spans open for more than one hour

yousif-bugsnag:
## Goal

Mitigates the potential for memory leaks caused by failing to end a span.

Currently if a span on the context stack is not ended for any reason (e.g. a developer has a problem with their logic for closing spans), then the span will remain open and there will be a reference to the span in the context stack until the thread / process / document (depending on platform) is killed.

## Design

Ideally we would use `WeakRef`s in the context stack to allow spans on the context stack to be garbage collected, however as `WeakRef` isn't supported on older browser versions we need an alternative approach.

So to avoid holding references indefinitely, any spans with a start time older than 1 hour are considered 'invalid', allowing them to be removed from the context stack naturally (since invalid spans are popped off the stack whenever a span is added or removed).

If/when an invalid span is ended, it will be discarded. As part of this, also fixed a bug where spans that were explicitly discarded using the special discard end time (-1) would still be considered valid and remain on the context stack.

## Testing

Added new unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Changed
+
+- (core) Discard spans open for more than one hour [#494](https://github.com/bugsnag/bugsnag-js-performance/pull/494)
+
 ## [v2.8.0] (2024-08-20)
 
 ### Added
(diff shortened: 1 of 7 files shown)


gingerbenw:
Also needs a changelog entry! 😄

tomlongridge (approved):

PR #496, opened 2024-08-30
Use 'r32' as the r value key in tracestate

imjoehaines:
## Goal

The JS SDK uses 32 bit values for p & r because JS numbers cannot hold 64 bit integers

We need to be able to determine if an r value is 32 or 64 bit in order to consume the tracestate correctly, so we're now using explicit bit sizes in the r value key (i.e. r32 & r64)


djskinner (approved):

imjoehaines:
Merging with CI failing as it's a known issue unrelated to this PR

PR #497, opened 2024-08-30
[PLAT-12625] Create new app start span wrapper

gingerbenw:
## Goal

Provide an alternative method to instrument app start spans aside from `setWrapperComponentProvider`

## Design

Using a higher order component (HOC) will prevent the app from re-rendering when root props are updated from the native side

## Changeset

- Expose new `withInstrumentedAppStarts` wrapper method from the performance client
- This PR also includes an alpha release for testing purposes

## Testing

- Add new end to end test to validate spans are still created using the new method
- Manual testing that the root app does not re-render when properties update from native iOS code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - (core) Discard spans open for more than one hour [#494](https://github.com/bugsnag/bugsnag-js-performance/pull/494)
 - (core) use API key subdomain as default endpoint [#500](https://github.com/bugsnag/bugsnag-js-performance/pull/500)
+- (react-native) Add new withInstrumentedAppStarts method to workaround automatically instrumented app start span issues [#497](https://github.com/bugsnag/bugsnag-js-performance/pull/497)
 
 ## [v2.8.0] (2024-08-20)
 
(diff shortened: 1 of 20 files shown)


yousif-bugsnag:
generally looks good to me - but I wonder if we should have some mechanism to prevent multiple app start spans from being created?

gingerbenw:
> I wonder if we should have some mechanism to prevent multiple app start spans from being created?

as in, if someone were to use the wrapper _and_ the `autoInstrumentAppStarts` config option?

yousif-bugsnag:
> as in, if someone were to use the wrapper and the autoInstrumentAppStarts config option?

Yeah - perhaps that's something we just need to document clearly, but seems like it could happen fairly easily because that config option is true by default.

yousif-bugsnag:
There's also nothing to stop someone wrapping multiple components using `withInstrumentedAppStarts` - they may even want to, for example an app that loads a login screen or goes straight to the main app screen depending on whether the user is authenticated might want to end the app start on whichever component is loaded first. But that could probably be a separate PR/future enhancement...

gingerbenw:
@yousif-bugsnag  added a simple check that only allows for a single app start span - create as many wrappers as you like 👍🏻

yousif-bugsnag (approved):

PR #498, opened 2024-08-30
Add React Native 0.75 to test matrix

yousif-bugsnag:
## Goal

Adds React Native 0.75 tests to the CI pipeline.

Also fixes the following issues with React Native test fixtures and e2e tests:

- Add pbxproj file and fix podspec in scenario launcher package (required for native iOS modules on older versions of React Native)
- Fix Android gradle file replacements for RN 0.64 (Flipper was not being fully removed, causing build issues on some CI machines and in local dev)
- Modify babel.config.js in test fixtures instead of replacing the whole file (Adds the required babel plugins while preserving the existing as this varies between react native versions)
- Fix typos in error correlation feature file

## Testing

Covered by a full CI run

### DIFF
--- a/bin/generate-react-native-fixture
+++ b/bin/generate-react-native-fixture
@@ -109,6 +109,15 @@ if (!process.env.SKIP_GENERATE_FIXTURE) {
     configureRN064Fixture(fixtureDir)
   }
 
+  // due to a bug in react-native-file-access, static frameworks is required on 0.75+ othwerwise iOS builds will fail.
+  // there is an open PR to fix this upstream: https://github.com/alpha0010/react-native-file-access/pull/88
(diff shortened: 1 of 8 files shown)


bin/generate-react-native-fixture
```diff
@@ -177,20 +187,20 @@ function configureRN064Fixture(fixtureDir) {
     nodeExecutableAndArgs: ["node", "--openssl-legacy-provider"]
   ]`
 
-  // remove flipper dependencies
+  // remove flipper and associated dependencies
```
gingerbenw, line 190: Why exactly do we do this, again? just to speed up the build process? what do we lose in the process?

gingerbenw (approved):

PR #499, opened 2024-09-03
[PLAT-12634] Make setAttribute public

gingerbenw:
## Goal

Support custom attributes

## Changeset

- Expose setAttribute method in the public span API
- Add span name validation

## Testing

Add unit test to ensure setAttribute adds attributes as necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


yousif-bugsnag:
Now that the API is public I think we should guard against non-typescript users inadvertently calling `setAttribute` with a non-string key.

I think spans already handle invalid attribute values when they're converted to JSON, but I think currently the code assumes the key will always be a string.

yousif-bugsnag (approved):

PR #500, opened 2024-09-03
use API key subdomain as default endpoint

djskinner:
## Goal

> The SaaS span endpoint now supports the format <project_api_key>.otlp.bugsnag.com
> This should just be the default value for the endpoint – if the developer passes their own URL we just use that value directly (don’t attempt to inject the API key into the provided URL).

## Design

Mutate the configuration so that the actual value used is available to all consumers of the configuration, such as request instrumentations which use the value to determine which requests should be tracked or not, and what URLs to ignore in request settling.

## Changeset

- use API key subdomain as default endpoint

## Testing

- [x] unit test
- [ ] manual test

Note: e2e tests not added the depend on using a custom endpoint, which prevents us from testing this functionality at the e2e level

### DIFF
--- a/packages/platforms/browser/tests/on-settle/index.test.ts
+++ b/packages/platforms/browser/tests/on-settle/index.test.ts
@@ -357,7 +357,7 @@ describe('onSettle', () => {
     expect(settleCallback).not.toHaveBeenCalled()
 
     // request should be ignored, so advancing by 100ms will settle
-    fetchRequestTracker.start({ ...START_CONTEXT, url: 'https://otlp.bugsnag.com/v1/traces' })
+    fetchRequestTracker.start({ ...START_CONTEXT, url: `https://${VALID_API_KEY}.otlp.bugsnag.com/v1/traces` })
 
     await jest.advanceTimersByTimeAsync(100)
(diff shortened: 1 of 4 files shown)


packages/platforms/browser/tests/on-settle/index.test.ts
```diff
@@ -357,7 +357,7 @@ describe('onSettle', () => {
     expect(settleCallback).not.toHaveBeenCalled()
 
     // request should be ignored, so advancing by 100ms will settle
-    fetchRequestTracker.start({ ...START_CONTEXT, url: 'https://otlp.bugsnag.com/v1/traces' })
+    fetchRequestTracker.start({ ...START_CONTEXT, url: `https://${VALID_API_KEY}.otlp.bugsnag.com/v1/traces` })
```
gingerbenw, line 360: without having dug too far into this test: are we hitting the trace server in a unit test?

djskinner: i don't think so, requestTracker doesn't do any sending and there is no "delivery", real or mock used in this test

gingerbenw (approved):

PR #501, opened 2024-09-03
[PLAT-12669] Handle array values in attributes

gingerbenw:


gingerbenw:
> Hey @Cawllec, I was hoping to test for empty array values using this step, but I assume it's sending `{}` as a string.
> 
> `| custom.array.empty | arrayValue | {} |`
> 
> Can you think of an easier way to test this, or should we update [this step](https://github.com/bugsnag/maze-runner/blob/c8cf41112e2fac8fa9cab1f9d81bd244d833c6dd/lib/features/steps/trace_steps.rb#L189)?

@Cawllec please ignore this one - we can just unit test this functionality

packages/core/lib/attributes.ts
```diff
@@ -53,13 +53,54 @@ export type ResourceAttributeSource<C extends Configuration>
...
   key: string
-  value: { stringValue: string }
-  | { intValue: string }
-  | { doubleValue: number }
-  | { boolValue: boolean }
+  value: { stringValue: string } | { intValue: string } | { doubleValue: number } | { boolValue: boolean }
 }
 
-export function attributeToJson (key: string, attribute: SpanAttribute): JsonAttribute | undefined {
+type AttributeArrayValueValue = Array<{ stringValue: string } | { intValue: string } | { doubleValue: number } | { boolValue: boolean }>
```
yousif-bugsnag, line 59: Should we be enforcing a single-type array here? maybe something like this?

```suggestion
type AttributeArrayValueValue = Array<{ stringValue: string }> | Array<{ intValue: string }> | Array<{ doubleValue: number }> | Array<{ boolValue: boolean }>
```

gingerbenw: actually no! those are perfectly valid - but when we _set_ them we try to enforce a single type - as per the `SpanAttribute` above

yousif-bugsnag: ah yeah of course - this is the internal type for the resulting JSON 🤦🏾

gingerbenw: I think I need to update the function and type names to make things a bit clearer

gingerbenw: made some updates - is this a little easier to read?

yousif-bugsnag: yeah that's a lot clearer - how about going a step further and making an `ArrayAttributeValue` interface?

```ts
interface StringAttributeValue { stringValue: string }
interface IntAttributeValue { intValue: string }
interface DoubleAttributeValue { doubleValue: number }
interface BoolAttributeValue { boolValue: boolean }

type SingleAttributeValue = StringAttributeValue | IntAttributeValue | DoubleAttributeValue | BoolAttributeValue

interface ArrayAttributeValue {
  arrayValue: {
    values?: SingleAttributeValue[]
  }
}

type JsonAttributeValue = SingleAttributeValue | ArrayAttributeValue

export interface JsonAttribute {
  key: string
  value: JsonAttributeValue
}

```

gingerbenw: Done and done!

yousif-bugsnag (approved):

PR #503, opened 2024-09-10
Refactor React Native App Start tests

hamudi-ahmed:
## Goal

Reworks the test fixture so that we can test app start instrumentation in a more realistic scenario

Previously the app start scenario runs after the app has already launched, which isn’t representative of a real app and caused us to miss a bug with initial background state.

## Design

The app start scenario now saves some static config via the native module and exits the app - when the app is restarted, the static config is read on launch and used to start the SDK immediately instead of starting the command runner.

Scenarios have also been refactored - instead of exporting a static config object, scenarios how have an initialise method which is called with some default configuration - scenarios can then extend/modify the config object as well as carry out any additional setup required for that specific test (e.g. save static config, start the notifier SDK, etc)

## Testing

Covered by CI


gingerbenw (approved):
👌🏻 much nicer!

PR #504, opened 2024-09-11
fix(core): avoid using URL class for apiKey subdomain

yousif-bugsnag:
## Goal

Avoid using `URL.hostname` to add the API key subdomain to the default endpoint.

URL properties such as hostname are not implemented in React Native and throw an error when accessed.

## Design

Replaced `URL` usage with a simple string replace to insert the API key - because the code only gets executed for the default endpoint the value is essentially hardcoded, and we have unit tests that would fail if the endpoint ever changed.

## Testing

This cannot be e2e tested as the e2e tests use a custom endpoint, so relied on existing unit tests and tested manually in a React Native app.


gingerbenw (approved):
if the tests are passing I'm happy 👍🏻

PR #514, opened 2024-10-01
Plat 12677/navigation span logic

gingerbenw:


gingerbenw:
@yousif-bugsnag interested to hear your thoughts on this change - not sure if it seems a bit excessive to publish a new package, but we currently have this logic replicated in 3 places

yousif-bugsnag:
My only question would be - what's the advantage of having this in a separate package rather than just exporting it from `@bugsnag/react-native-performance`? The two navigation plugins that use it already have a peer dependency on `@bugsnag/react-native-performance`

gingerbenw:
@yousif-bugsnag that. is a much better idea.

package-lock.json
yousif-bugsnag, line 1: This still has the non-existent packages/create-navigation-span listed

yousif-bugsnag (approved):
lockfile needs fixing but otherwise LGTM

package-lock.json
gingerbenw: whoops! thanks. 🙇🏻

PR #516, opened 2024-10-03
Add buildkite test analytics

Cawllec:


twometresteve (approved):
Change looks fine and maybe environment variable name length isn't an issue at all, I just wonder if they could be shortened in general with the following?
- REACT_NATIVE -> RN
- PERFORMANCE -> PERF
- BUILDKITE_ANALYTICS -> TEST_ANALYTICS

Only and idea though, feel free to ignore.

PR #517, opened 2024-10-04
Remove CI pipeline Docker env files

twometresteve:
## Goal

Fix the following overnight failures building the Maze Runner image:
```
Failed to load /var/lib/buildkite-agent/builds/docker_env_file: open /var/lib/buildkite-agent/builds/docker_env_file: no such file or directory
```

## Design

It's unclear how the failures started (why best guess is an S3 secrets change), but we don't actually provide a Docker env file anyway and so this complexity can just be removed.

## Testing

Covered by a full CI run.


yousif-bugsnag (approved):
Seems fine to me

PR #519, opened 2024-10-16
RN native integration e2e test setup

yousif-bugsnag:
## Goal

e2e test setup for native performance integration in React Native:

 - Generate a separate e2e test fixture with native performance SDKs installed.
 - Add support for starting the native Android and Cocoa SDKs through the scenario launcher
 - Add a basic scenario that sends a custom native span - this is really a placeholder test to make sure that the setup works and will be replaced when we start to add functionality.

## Design

A separate test fixture is required so that we can make sure the SDK continues to work as expected when native Android and Cocoa performance SDKs are not installed (existing fixtures and tests).

The test fixture generation script will only install the native Android and Cocoa SDKs when the `NATIVE_INTEGRATION` environment variable is set.

## Testing

Covered by CI

### DIFF
--- a/.buildkite/react-native-pipeline.yml
+++ b/.buildkite/react-native-pipeline.yml
@@ -297,6 +297,7 @@ steps:
               - --appium-version=1.22
               - --no-tunnel
               - --aws-public-ip
+              - --tags=@native_integration
           test-collector#v1.10.2:
             files: "reports/TEST-*.xml"
             format: "junit"
@@ -333,6 +334,7 @@ steps:
               - --appium-version=1.22
               - --no-tunnel
               - --aws-public-ip
+              - --tags=@native_integration
           test-collector#v1.10.2:
             files: "reports/TEST-*.xml"
             format: "junit"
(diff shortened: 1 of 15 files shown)


.buildkite/react-native-pipeline.yml
```diff
@@ -55,6 +55,59 @@ steps:
...
+          RN_VERSION: "{{matrix}}"
+          NOTIFIER_VERSION: '8.0.0'
+          BUILD_ANDROID: "true"
+          NATIVE_INTEGRATION: "1"
+        artifact_paths:
+          - "test/react-native/features/fixtures/generated/native-integration/old-arch/**/reactnative.apk"
+        plugins:
+          artifacts#v1.9.3:
+            upload: "test/react-native/features/fixtures/generated/native-integration/old-arch"
+            compressed: rn-projects.zip
```
gingerbenw, line 75: is this just here for debugging? these are pretty big bundles to upload on every build otherwise

gingerbenw (approved):

.buildkite/react-native-pipeline.yml
yousif-bugsnag: ah - this was copied inadevertently from another step - I think it was added in the error correlation PR but I guess it's not needed there either. I'll remove it from both 👍🏾

PR #525, opened 2024-11-07
Add support for attaching to native performance SDKs

yousif-bugsnag:
## Goal

Adds a new `BugsnagPerformance.attach` method to React Native Performance to allow attaching to an already running native (Android or Cocoa) performance SDK.

This is effectively a wrapper around `BugsnagPerformance.start` that loads the native configuration and uses this to start the native SDK

## Design

Added a new method `getNativeConfiguration` to the Turbo Module, which returns the finalised configuration passed to native `start` or `null` if the native SDK has not been started.

When `attach` is called, the native config is merged with JS-specific config passed to `attach` before calling `start`

`BugsnagPerformance.attach` will log a warning and noop if the native SDK is not installed or has not been started.

## Testing

- Added new unit tests for attach
- Updated the e2e tests to check that the React Native SDK is started using native config and a span is delivered. 
- Updated the test fixture generation to install Cocoa and Android Performance from integration branches

### DIFF
--- a/packages/platforms/react-native/android/build.gradle
+++ b/packages/platforms/react-native/android/build.gradle
@@ -36,5 +36,5 @@ repositories {
 
 dependencies {
   implementation 'com.facebook.react:react-native'
-  compileOnly("com.bugsnag:bugsnag-android-performance:+")
+  compileOnly("com.bugsnag:bugsnag-android-performance:9.9.9")
 }
--- a/packages/platforms/react-native/android/build.gradle
+++ b/packages/platforms/react-native/android/build.gradle
@@ -36,4 +36,5 @@ repositories {
 
 dependencies {
   implementation 'com.facebook.react:react-native'
+  compileOnly("com.bugsnag:bugsnag-android-performance:+")
 }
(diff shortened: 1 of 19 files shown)


packages/platforms/react-native/lib/NativeBugsnagPerformance.ts
```diff
@@ -10,11 +10,26 @@ export type DeviceInfo = {
   bundleIdentifier?: string
 }
 
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
gingerbenw, line 13: what needed disabling here? interface vs type?

gingerbenw (approved):

packages/platforms/react-native/android/build.gradle
```diff
@@ -36,4 +36,5 @@ repositories {
 
 dependencies {
   implementation 'com.facebook.react:react-native'
+  compileOnly("com.bugsnag:bugsnag-android-performance:+")
```
lemnik, line 39: given that the `generate-react-native-fixture` defined the published version number, we could rather specify that here as well (and avoid possibly picking up a "real" version by accident)?

```suggestion
  compileOnly("com.bugsnag:bugsnag-android-performance:9.9.9")
```

lemnik (approved):
LGTM - with a minor suggesion

packages/platforms/react-native/lib/NativeBugsnagPerformance.ts
yousif-bugsnag: yeah, IIRC using interfaces causes issues with React Native's codegen (at least on some versions)

PR #530, opened 2024-11-19
Replace FileSystem wrapper object with functions from our turbomodule

DariaKunoichi:
## Goal
Replace `react-native-file-access` dependency with our own turbomodule's functions.

## Changeset
Added functions to existing turbomodule interface.
Part of native module's methods extracted to FileSystem object which has the same interface as removed module.
Also added a check to ensure native module is not null, with either old or new architecture.

## Testing
For unit tests methods had to be added to `react-native` mock, including the mock for `NativeModules`.


packages/platforms/react-native/lib/NativeBugsnagPerformance.ts
```diff
@@ -1,6 +1,8 @@
 import type { TurboModule } from 'react-native/Libraries/TurboModule/RCTExport'
 import { TurboModuleRegistry } from 'react-native'
 
+export type Encoding = 'utf8' | 'base64'
```
yousif-bugsnag, line 4: I think this is what's causing the `Union types are unsupported in structs` codegen error.

We can use this type on the JS side, but in the turbo module spec this should just be a `string`

packages/platforms/react-native/lib/persistence/file-native.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const LINKING_ERROR =
+  'The package \'BugsnagReactNativePerformance\' doesn\'t seem to be linked.'
+
+// @ts-expect-error check if turbomodule is enabled
+const isTurboModuleEnabled = global.__turboModuleProxy != null
+
+const BugsnagFileAccessModule = isTurboModuleEnabled
+  ? require('./NativeBugsnagPerformance').default
+  : NativeModules.BugsnagReactNativePerformance
```
yousif-bugsnag, line 12: this shouldn't be necessary - there's a file called `native.ts` that handles this logic, so you can just do:

```ts
import NativeBugsnagPerformance from '../native'
```

DariaKunoichi: removed

yousif-bugsnag (approved):

PR #532, opened 2024-11-21
add sendPayloadChecksums config option and implement Bugsnag-Integrit…

djskinner:
…y header


packages/core/lib/core.ts
```diff
@@ -73,6 +73,11 @@ export function createClient<S extends CoreSchema, C extends Configuration, T> (
     start: (config: C | string) => {
       const configuration = validateConfig<S, C>(config, options.schema)
 
+      // sendPayloadChecksums is false by default unless custom endpoints are not specified
+      if (typeof config !== 'string' && !config.endpoint) {
+        configuration.sendPayloadChecksums = ('sendPayloadChecksums' in config && config.sendPayloadChecksums) || true
+      }
```
djskinner, line 79: Not keen on this but not sure of a better way. We need to work out what config options the user passed in to determine the behaviour. `const configuration = validateConfig<S, C>(config, options.schema)` does more than just validate, it creates a config with all values set based on the schema, using default values if not supplied.

.buildkite/browser-pipeline.full.yml
```diff
@@ -29,7 +29,6 @@ steps:
         run: browser-maze-runner-bs
         use-aliases: true
         command:
-          - --https
```
djskinner, line 32: i noticed this wasn't working for the ios mobile devices (but it does pass for anrdoid). Wasn't sure it was worth splitting this so that we run over https on android so ive just disabled https for all mobile device runs

packages/core/lib/core.ts
gingerbenw: right now this is the place we're performing similar manipulation on the configuration, perhaps we could refactor the validation to include this, possibly separating out validation and 'cleaning' but I think for now this is fine

packages/delivery-fetch/tests/delivery.test.ts
```diff
@@ -293,10 +298,117 @@ describe('Browser Delivery', () => {
...
+        }]
+      },
+      headers: {
+        'Bugsnag-Api-Key': 'test-api-key',
+        'Content-Type': 'application/json',
+        'Bugsnag-Span-Sampling': '1:1'
+      }
+    }
+
+    window.isSecureContext = false
```
gingerbenw, line 335: does this remain `false` for the next test? we're using `beforeAll` rather than `beforeEach` to revert this back to true

djskinner: Ah yeah 👍 fixed

gingerbenw (approved):
looks good to me - a rebase from next should resolve the ios https tests

PR #533, opened 2024-11-22
run e2e tests over https

djskinner:
## Goal

Get the browser e2e tests running over https so we can test functionality that requires a secure context.

## Design/Changeset

- Set the new --https flag on maze runner
- Update the test fixtures to use the new docs endpoint being served from `/docs`

## Testing

e2e tests passing locally and on CI


test/browser/features/lib/url-generator.rb
```diff
@@ -5,7 +5,7 @@ def initialize(maze_uri, fixtures_uri)
   end
 
   def for_path(path)
-    uri = URI.join(@fixtures_uri, path)
+    uri = URI.join(@maze_uri, "/docs#{path}")
```
djskinner, line 8: Prefixing with docs here allows: `Given I navigate to the test URL "/vue-router"` 

However, the text fixture and assertions still need to be well aware they're actually being served from `/docs/vue-router`. So I wonder if it would be better not to do the prefix and instead require `Given I navigate to the test URL "/docs/vue-router"`. Potentially seems a bit clearer?

test/browser/features/page-load-spans.feature
```diff
@@ -6,113 +6,113 @@ Feature: Page Load spans
         And I click the element "stop-clock"
         And I wait to receive 1 trace
 
-        Then a span named "[FullPageLoad]/page-load-spans/" contains the attributes:
-            | attribute                         | type             | value                     |
-            | bugsnag.span.category             | stringValue      | full_page_load            |
-            | bugsnag.browser.page.title        | stringValue      | New title                 |
-            | bugsnag.browser.page.route        | stringValue      | /page-load-spans/         |
+        Then a span named "[FullPageLoad]/docs/page-load-spans/" contains the attributes:
```
djskinner, line 9: Example of the above comment here. `"Given I navigate to the test URL "/page-load-spans"` yet I end up with a `"span named [FullPageLoad]/docs/page-load-spans/"`

test/browser/features/lib/url-generator.rb
gingerbenw: I agree - including `/docs` in all steps would definitely be clearer. I would prefer to go for that approach.

test/browser/Gemfile.lock
twometresteve, line 1: We just remove Gemfile.lock from config control now to reduce Dependabot alerts.

test/browser/features/lib/url-generator.rb
twometresteve: Ah yes, I see your perspective on this now.  At first, naively, `/docs` seemed more like an incidental detail of the test harness to me, but now I see why it matters in the context of browsers and URLs.

twometresteve (commented):
Changes so far look good.  Do we know why the Safari tests are failing?  Also I would run a full pipeline before merging the PR.

djskinner:
> Do we know why the Safari tests are failing?

Looks like a problem running the tests over https: `ERROR: SSL_accept SYSCALL returned=5 errno=0 state=SSLv3/TLS write server done` - I'll configure the pipeline not to run these ones over https

test/browser/features/support/maze-config.rb
```diff
@@ -16,9 +16,15 @@ def get_test_url()
     maze_address = "#{host}:9339"
   end
 
+  if Maze.config.https
+    protocol = 'https'
+  else
+    protocol = 'http'
+  end
+
```
twometresteve, line 24: Ruby has the ternary operator, although I'm not suggesting you suffer another build to change it, just do it on another branch.
```suggestion
  protocol = Maze.config.https ? 'https' : 'http'
```

twometresteve (approved):

PR #536, opened 2024-11-28
feat(react-native): support for starting native spans on android

yousif-bugsnag:
## Goal

Add support for starting native spans on Android

## Design

Added a new Turbo Module method called `startNativeSpan` which creates a span in the Android Performance SDK and returns a JS representation of the native Android span to JS.

The Turbo Module method accepts a limited set of span options (`startTime` and `parentContext`) as native spans should always be first class and should never become the current context in the native SDK.

## Testing

Tested manually as this is not yet being used

### DIFF
--- a/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
+++ b/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
@@ -9,10 +9,20 @@ import androidx.annotation.Nullable;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
-import com.facebook.react.bridge.WritableArray;
+import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import java.security.SecureRandom;
+import java.util.UUID;
+import java.util.concurrent.Callable;
+
 import com.bugsnag.android.performance.BugsnagPerformance;
+import com.bugsnag.android.performance.SpanContext;
+import com.bugsnag.android.performance.SpanOptions;
+
+import com.bugsnag.android.performance.internal.BugsnagClock;
+import com.bugsnag.android.performance.internal.EncodingUtils;
+import com.bugsnag.android.performance.internal.SpanFactory;
+import com.bugsnag.android.performance.internal.SpanImpl;
 import com.bugsnag.android.performance.internal.processing.ImmutableConfig;
 
 class NativeBugsnagPerformanceImpl {
@@ -111,7 +121,7 @@ class NativeBugsnagPerformanceImpl {
     result.putString("releaseStage", nativeConfig.getReleaseStage());
     result.putString("serviceName", nativeConfig.getServiceName());
     result.putInt("attributeCountLimit", nativeConfig.getAttributeCountLimit());
-    result.putInt("attriubuteStringValueLimit", nativeConfig.getAttributeStringValueLimit());
+    result.putInt("attributeStringValueLimit", nativeConfig.getAttributeStringValueLimit());
     result.putInt("attributeArrayLengthLimit", nativeConfig.getAttributeArrayLengthLimit());
 
     var appVersion = nativeConfig.getAppVersion();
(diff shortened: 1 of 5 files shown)


gingerbenw (approved):

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -130,9 +140,98 @@ public WritableMap getNativeConfiguration() {
...
+        @Override
+        public Runnable wrap(@NonNull Runnable runnable) {
+            return null;
+        }
+
+        @NonNull
+        @Override
+        public <T> Callable<T> wrap(@NonNull Callable<T> callable) {
+            return null;
+        }
```
lemnik, line 228:
```suggestion
        @NonNull
        @Override
        public Runnable wrap(@NonNull Runnable runnable) {
            return SpanContext.DefaultImpls.wrap(this, runnable);
        }

        @NonNull
        @Override
        public <T> Callable<T> wrap(@NonNull Callable<T> callable) {
            return SpanContext.DefaultImpls.wrap(this, callable);
        }
```

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -130,9 +140,98 @@ public WritableMap getNativeConfiguration() {
...
+  }
+
+  private SpanContext jsSpanContextToNativeSpanContext(ReadableMap spanContext) {
+    String spanId = spanContext.getString("id");
+    String traceId = spanContext.getString("traceId");
+
+    long nativeSpanId = Long.parseUnsignedLong(spanId, 16);
+    UUID nativeTraceId = new UUID(Long.parseUnsignedLong(traceId.substring(0, 16), 16), Long.parseUnsignedLong(traceId.substring(16), 16));
+
+    SpanContext nativeSpanContext = new SpanContext() {
```
lemnik, line 206: Any reason this shouldn't maybe be a named class (like `ReactNativeSpanContext`) to remove the noise from this `jsSpanContextToNativeSpanContext`?

yousif-bugsnag: good idea 👍🏾

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -130,9 +135,67 @@ public WritableMap getNativeConfiguration() {
...
+    }
+
+    return span;
+  }
+
+  private SpanOptions readableMapToSpanOptions(ReadableMap jsOptions) {
+    SpanOptions spanOptions = SpanOptions.DEFAULTS
+      .setFirstClass(true)
+      .makeCurrentContext(false)
+      .within(null);
```
lemnik, line 177: This can be safely moved into a `private static final` field:

```
private static final SpanOptions BRIDGED_SPAN_OPTIONS = SpanOptions
    .createFirstClass(true)
    .makeCurrentContext(false)
    .within(null);
```

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/ReactNativeSpanContext.java
```diff
@@ -0,0 +1,42 @@
...
+class ReactNativeSpanContext implements SpanContext {
+  private final long nativeSpanId;
+  private final UUID nativeTraceId;
+
+  public ReactNativeSpanContext(String spanId, String traceId) {
+    nativeSpanId = Long.parseUnsignedLong(spanId, 16);
+    nativeTraceId = new UUID(
+      Long.parseUnsignedLong(traceId.substring(0, 16), 16),
+      Long.parseUnsignedLong(traceId.substring(16), 16)
+    );
```
lemnik, line 17: It feels like `16` for the radix should maybe be made into a constant?
```java
private static final int HEX_RADIX = 16;
```

It might also be worth it to have a `TRACE_ID_MIDPOINT = 16` constant

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/ReactNativeSpanContext.java
```diff
@@ -0,0 +1,42 @@
...
+import androidx.annotation.NonNull;
+import com.bugsnag.android.performance.SpanContext;
+import java.util.UUID;
+import java.util.concurrent.Callable;
+
+class ReactNativeSpanContext implements SpanContext {
+  private final long nativeSpanId;
+  private final UUID nativeTraceId;
+
+  public ReactNativeSpanContext(String spanId, String traceId) {
```
lemnik, line 12: nit: visibility

```suggestion
 ReactNativeSpanContext(String spanId, String traceId) {
```

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
yousif-bugsnag: I don't think we can do this - it will crash if `bugsnag-android-performance` is not installed

lemnik (approved):

PR #537, opened 2024-11-28
Add android implementation for FileSystem operations

DariaKunoichi:
## Goal
Replace react-native-file-access module with our own implementation for native file io.

## Changeset
Added implementation for the android native part.

## Testing
Used existing mazerunner tests.

### DIFF
--- a/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
+++ b/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
@@ -32,7 +32,7 @@ class NativeBugsnagPerformanceImpl {
     this.reactContext = reactContext;
   }
 
-  public WritableMap getDeviceInfo() {
+  WritableMap getDeviceInfo() {
     WritableMap map = Arguments.createMap();
     try {
       String bundleIdentifier = this.reactContext.getPackageName();
@@ -62,7 +62,7 @@ class NativeBugsnagPerformanceImpl {
     return map;
   }
 
-  public String requestEntropy() {
+  String requestEntropy() {
     byte[] bytes = new byte[1024];
     random.nextBytes(bytes);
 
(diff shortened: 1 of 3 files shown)


packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
lemnik, line 1: I don't think any of the methods in this class need to be `public`

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -74,6 +81,101 @@ public void requestEntropyAsync(Promise promise) {
...
+  public void readFile(String path, String encoding, Promise promise) {
+    try {
+      StringBuilder fileContent = new StringBuilder();
+      InputStreamReader isr = new InputStreamReader(new FileInputStream(path), encoding);
+      BufferedReader fileReader = new BufferedReader(isr);
+      String dataLine = "";
+      while ((dataLine = fileReader.readLine()) != null) {
+        fileContent.append(dataLine);
+      }
+      fileReader.close();
```
lemnik, line 148:
```suggestion
    File file = new File(path);
    StringBuilder fileContent = new StringBuilder((int) file.length());
    try(
      FileInputStream fin = new FileInputStream(file);
      InputStreamReader isr = new InputStreamReader(fin, encoding);
    ) {
      char[] buffer = new char[4096];
      int charsRead = 0;
      while ((charsRead = isr.read(buffer)) != -1) {
        fileContent.append(buffer, 0, charsRead);
      }
```

I think using a fixed sized buffer of characters and pre-allocating the `StringBuilder` is going to be better than reading the lines one-by-one.

Also using `try-with-resources` will handle close operations for us.

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -74,6 +81,101 @@ public void requestEntropyAsync(Promise promise) {
...
+    } catch(Exception e) {
+      promise.reject(e);
+    }
+  }
+
+  public void writeFile(String path, String data, String encoding, Promise promise){
+    try {
+      Writer w = new OutputStreamWriter(new FileOutputStream(path), encoding);
+      w.write(data);
+      w.close();
```
lemnik, line 172:
```suggestion
  public void writeFile(String path, String data, String encoding, Promise promise){
    try(
      FileOutputStream fout = new FileOutputStream(path);
      Writer w = new OutputStreamWriter(fout, encoding);
    ) {
      w.write(data);
```

PR #539, opened 2024-11-29
Implement current state enum in client

AnastasiiaSvietlova:
## Changeset

Add a new `appState` property and `setAppState` to the core client.

## Testing

Add unit test to test `setAppState` 
Update `createTestClient` default options

### DIFF
--- a/packages/test-utilities/lib/create-test-client.ts
+++ b/packages/test-utilities/lib/create-test-client.ts
@@ -15,7 +15,7 @@ import {
 
 } from '@bugsnag/core-performance'
 import type { BugsnagPerformance, ClientOptions, Configuration, CoreSchema, Delivery } from '@bugsnag/core-performance'
-import { AppState } from '@bugsnag/react-native-performance/__mocks__/react-native'
+import type { AppState } from '../../core'
 
 const defaultOptions = () => ({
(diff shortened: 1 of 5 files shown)


gingerbenw (approved):
Looks good - but could we create a new `integration/component-rendering` branch and target that please? 🙏🏻

packages/test-utilities/lib/create-test-client.ts
```diff
@@ -25,6 +26,8 @@ const defaultOptions = () => ({
   spanAttributesSource,
   schema,
   plugins: () => [],
+  appState: 'starting',
+  setAppState: () => AppState,
```
gingerbenw, line 30:
```suggestion
  setAppState: (appState: AppState) => void,
```

gingerbenw (changes requested):
Sorry, just spotted another change

packages/test-utilities/lib/create-test-client.ts
AnastasiiaSvietlova: In this case `'appState' is declared but its value is never read`
Maybe:  `setAppState: (appState: AppState) => {appState} `?

gingerbenw (approved):

PR #540, opened 2024-11-29
feat(react-native): support for starting native spans on ios

yousif-bugsnag:
## Goal

Add support for starting native spans on iOS - this is the iOS implementation of the `startNativeSpan` Turbo Module method added in #536 

#### Start time conversions
In Cocoa Performance span start/end times are specified as `NSDate` objects (converted to `CFAbsoluteTime` under the hood), so start times from the React Native clock (using `performance.now()`) need to be sent as unix timestamps over the wire and converted to/from `NSDate` objects in the turbo module. 

#### Turbo Module types
Updated the Turbo Module Spec to use the `UnsafeObject` type for native span options - this is required for backwards compatibility with the Old Architecture (otherwise codegen will generate a struct for the typed function parameter instead of using `NSDictionary`)
 
## Testing

Tested manually as this isn't used yet


packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -129,6 +138,29 @@ @implementation BugsnagReactNativePerformance
...
+    spanOptions[@"firstClass"] = [NSNumber numberWithBool:YES];
+    
+    BugsnagPerformanceSpan *nativeSpan = [BugsnagCocoaPerformanceFromBugsnagReactNativePerformance.sharedInstance startSpan:name options:spanOptions];
+    [nativeSpan.attributes removeAllObjects];
+
+    NSMutableDictionary *span = [NSMutableDictionary new];
+    span[@"name"] = nativeSpan.name;
+    span[@"id"] = [NSString stringWithFormat:@"%llx", nativeSpan.spanId];
+    span[@"traceId"] = [NSString stringWithFormat:@"%llx%llx", nativeSpan.traceId.hi, nativeSpan.traceId.lo];
+    span[@"startTime"] = [NSNumber numberWithDouble: [nativeSpan.startTime timeIntervalSince1970] * NSEC_PER_SEC];
```
gingerbenw, line 156: do we also need to add the first class property, or does that only matter for the native side?

bin/generate-react-native-fixture
```diff
@@ -381,7 +381,7 @@ function installCocoaPerformance() {
   const podfilePath = resolve(fixtureDir, 'ios/Podfile')
   let podfile = fs.readFileSync(podfilePath, 'utf8')
 
-  const performancePod = `pod 'BugsnagPerformance', :git => 'https://github.com/bugsnag/bugsnag-cocoa-performance.git', :branch => 'integration/react-native-integration'`
+  const performancePod = `pod 'BugsnagPerformance', :git => 'https://github.com/bugsnag/bugsnag-cocoa-performance.git', :branch => 'ya/crosstalk-api-startspan-improvements'`
```
gingerbenw, line 384: I'm sure you won't miss it, but just a reminder to update this branch!

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
yousif-bugsnag: The first class span option is always true for native spans (set a bit further up), but the actual attribute will be set in JS (the JS code that calls this hasn't been written yet)

kstenerud (approved):

PR #545, opened 2024-12-03
Add ios implementation for FileSystem operations

DariaKunoichi:
## Goal
Replace `react-native-file-access` module with our implementation for file I/O.

## Changeset
Added implementation for filesystem methods for IOS.

## Testing
Used existing maze-runner tests.

### DIFF
--- a/packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
+++ b/packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
@@ -90,8 +90,15 @@ RCT_EXPORT_METHOD(requestEntropyAsync:(RCTPromiseResolveBlock)resolve
 
 RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getNativeConstants) {
     NSMutableDictionary *nativeDirs = [NSMutableDictionary new];
-    nativeDirs[@"CacheDir"] = NSSearchPathForDirectoriesInDomains(.cachesDirectory, .userDomainMask, true).first!;
-    nativeDirs[@"DocumentDir"] = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first!;
+    NSArray<NSString *> *caches = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, true);
+    if (caches != nil && [caches count] != 0 ) {
+        nativeDirs[@"CacheDir"] = caches[0];
+    }
+    NSArray<NSString *> *documents = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, true);
+    if (documents != nil && [documents count] != 0 ) {
+        nativeDirs[@"DocumentDir"] = documents[0];
+    }
+
     return nativeDirs;
 }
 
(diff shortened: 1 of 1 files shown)


packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -82,12 +82,118 @@ @implementation BugsnagReactNativePerformance
...
- }
+RCT_EXPORT_METHOD(requestEntropyAsync:(RCTPromiseResolveBlock)resolve
+                   rejecter:(RCTPromiseRejectBlock)reject) {
+    NSString *hexStr = getRandomBytes();
+    resolve(hexStr);
+}
+
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getNativeConstants) {
+    NSMutableDictionary *nativeDirs = [NSMutableDictionary new];
+    nativeDirs[@"CacheDir"] = NSSearchPathForDirectoriesInDomains(.cachesDirectory, .userDomainMask, true).first!;
```
robert-smartbear, line 93: Even though we don't expect the result to be empty, I think that adding a null check instead of a force unwrap would be a better option here

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -82,12 +82,118 @@ @implementation BugsnagReactNativePerformance
...
+    nativeDirs[@"DocumentDir"] = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first!;
+    return nativeDirs;
+}
+
+RCT_EXPORT_METHOD(exists:(NSString *)path
```
robert-smartbear, line 98: I would name the method `fileExistsAtPath`: instead of just `exists:`

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -82,12 +82,118 @@ @implementation BugsnagReactNativePerformance
...
+
+RCT_EXPORT_METHOD(isDir:(NSString *)path
+      resolve:(RCTPromiseResolveBlock)resolve
+       reject:(RCTPromiseRejectBlock)reject) {
+    BOOL* isDir;
+    BOOL exists = [NSFileManager.defaultManager fileExistsAtPath:path isDirectory:isDir];
+    resolve(@(exists && isDir));
+}
+
+RCT_EXPORT_METHOD(ls:(NSString *)path
```
robert-smartbear, line 113: I propose a more descriptive name like `contentsOfDirectoryAtPath:`

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -82,12 +82,118 @@ @implementation BugsnagReactNativePerformance
...
-     NSString *hexStr = getRandomBytes();
-     resolve(hexStr);
- }
+RCT_EXPORT_METHOD(requestEntropyAsync:(RCTPromiseResolveBlock)resolve
+                   rejecter:(RCTPromiseRejectBlock)reject) {
+    NSString *hexStr = getRandomBytes();
+    resolve(hexStr);
+}
+
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getNativeConstants) {
```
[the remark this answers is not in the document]
DariaKunoichi: It's actually a temporary method - it's supposed to be getConstants which uses react native special method for delivering static constants from native side. 
But for now I am using this method. It's not having resolve/reject because it's synchronous.

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
DariaKunoichi: changed

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
DariaKunoichi: it's the same name as in replaced module - I wanted to have same interface

robert-smartbear (approved):

PR #548, opened 2024-12-16
feat(react-native): support ending native android spans

yousif-bugsnag:
## Goal

Update the Android Turbo Module to support ending native spans, including support for marking a native span's end time independently of processing the span.

## Design

Added two new Turbo Module methods:
- `markNativeSpanEndTime` - sets the end time on a span without sending the span to be processed by the native SDK. This allows us to mark the end time and get an accurate end metrics snapshot in situations where the end time is overridden, e.g. navigation plugins.

- `endNativeSpan` - fully ends a native span and sends it for processing by the native SDK, translating the final JS attributes into native span attributes.

The Turbo Module now stores references to native spans in a `HashMap` when they are started so that they can be retrieved and ended later. The reference is removed when the span is ended.

Added a new `ReactNativeSpanAttributes` class to handle conversion of JS span attributes from a `ReadableMap` to native span attributes. Since all numbers passed from JS are doubles, we need to iterate through each attribute and determine it's type, and for double attributes we need to determine if it is actually a double or an integer.

## Testing

Tested manually as this isn't called from JS yet

### DIFF
--- a/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
+++ b/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
@@ -1,15 +1,18 @@
 package com.bugsnag.reactnative.performance;
 
+import android.annotation.SuppressLint;
 import android.content.pm.PackageInfo;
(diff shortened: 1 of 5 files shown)


lemnik (approved):
LGTM

PR #553, opened 2024-12-19
update current state in page load and navigation

AnastasiiaSvietlova:

### DIFF
--- a/packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
+++ b/packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
@@ -4,6 +4,7 @@ import type { BrowserConfiguration } from '../config'
 import type { RouteChangeSpanEndOptions, RouteChangeSpanOptions } from '../routing-provider'
 import { getPermittedAttributes } from '../send-page-attributes'
 import { defaultRouteResolver } from '../default-routing-provider'
+import type { AppState } from '../../../../core/lib/core'
 
 // exclude isFirstClass from the route change option schema
 const { startTime, parentContext, makeCurrentContext } = coreSpanOptionSchema
@@ -27,7 +28,8 @@ export class RouteChangePlugin implements Plugin<BrowserConfiguration> {
   constructor (
     private readonly spanFactory: SpanFactory<BrowserConfiguration>,
     private readonly location: Location,
-    private readonly document: Document
+    private readonly document: Document,
+    private readonly setAppState: (appState: AppState) => void
   ) {}
 
   configure (configuration: InternalConfiguration<BrowserConfiguration>) {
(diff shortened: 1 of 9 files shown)


packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
```diff
@@ -4,6 +4,7 @@ import type { BrowserConfiguration } from '../config'
 import type { RouteChangeSpanEndOptions, RouteChangeSpanOptions } from '../routing-provider'
 import { getPermittedAttributes } from '../send-page-attributes'
 import { defaultRouteResolver } from '../default-routing-provider'
+import type { AppState } from '../../../../core/lib/core'
```
gingerbenw, line 7: This should be an import from `@bugsnag/core` and not a relative import

packages/platforms/browser/tests/auto-instrumentation/route-change-plugin.test.ts
```diff
@@ -10,6 +10,7 @@ import type { BrowserConfiguration, BrowserSchema } from '../../lib/config'
 import { createDefaultRoutingProvider } from '../../lib/default-routing-provider'
 import type { OnSettle } from '../../lib/on-settle'
 import type { StartRouteChangeCallback } from '../../lib/routing-provider'
+import type { AppState } from '../../../../core/lib/core'
```
gingerbenw, line 13: same as above.

packages/platforms/browser/lib/auto-instrumentation/route-change-plugin.ts
AnastasiiaSvietlova: So this import should be fixed in all plugins?

gingerbenw: yep, if we need to use code from another library, it should be exported and imported appropriately.

packages/platforms/browser/tests/auto-instrumentation/full-page-load-plugin.test.ts
```diff
@@ -39,6 +40,10 @@ describe('FullPageLoadPlugin', () => {
     performance.addEntry(createPerformancePaintTimingFake({ startTime: 128 }))
     performance.addEntry(createPerformanceEventTimingFake({ startTime: 0.4, processingStart: 1 }))
 
+    let appState: AppState = 'starting'
+    const setAppState = jest.fn((state: AppState) => {
+      appState = state
+    })
```
gingerbenw, line 46: are you not able to get these from the `plugins: (spanFactory)` of the test client on line 57? they should be part of the base client so no need to mock them here

yousif-bugsnag (approved):
LGTM

PR #556, opened 2024-12-31
Delegate first class spans to native SDK when attached

yousif-bugsnag:
## Goal

Delegates the creation and processing of first class spans in React Native to the native Android/Cocoa Performance SDK when attached.

## Design

If a native SDK is running and attached (by calling `BugsnagPerformance.attach`), first-class spans are started, processed and discarded in the native SDK.

JS spans that are backed by a native span are wrapped in a `NativeSpanInternal` class to distinguish them from regular JS spans. 

Refactored the core start logic so that the span factory is configured synchronously as soon as start is called - this is so that `onSpanEnd` callbacks can be applied to spans started immediately after `attach` is called. Note that any first class spans started before `attach` is called will not be delegated and will be started/processed in JS.

## Testing
- Added unit tests for ReactNativeSpanFactory class
- Updated native integration e2e test scenario to verify that native spans are processed correctly

### DIFF
--- a/packages/core/lib/span-factory.ts
+++ b/packages/core/lib/span-factory.ts
@@ -6,7 +6,7 @@ import type { Configuration, InternalConfiguration, Logger, OnSpanEndCallbacks }
 import { defaultSpanAttributeLimits } from './custom-attribute-limits'
 import type { IdGenerator } from './id-generator'
 import type { NetworkSpanOptions } from './network-span'
-import type { Processor } from './processor'
+import type { BufferingProcessor, Processor } from './processor'
 import type { ReadonlySampler } from './sampler'
 import type { InternalSpanOptions, ParentContext, Span, SpanOptionSchema, SpanOptions } from './span'
 import { SpanInternal, coreSpanOptionSchema } from './span'
@@ -117,8 +117,7 @@ export class SpanFactory<C extends Configuration> {
     return spanInternal
   }
 
-  configure (processor: Processor, configuration: InternalConfiguration<C>) {
-    this.processor = processor
+  configure (configuration: InternalConfiguration<C>) {
     this.logger = configuration.logger
     this.spanAttributeLimits = {
       attributeArrayLengthLimit: configuration.attributeArrayLengthLimit,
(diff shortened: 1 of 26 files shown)


packages/platforms/react-native/tests/span-factory.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    it('does not discard a non-native span', () => {
+      const startTime = clock.now()
+      const span = spanFactory.startSpan('not native', { startTime, isFirstClass: false })
+      expect(NativeBugsnagPerformance!.startNativeSpan).not.toHaveBeenCalled()
+
+      spanFactory.endSpan(span, clock.now())
+      expect(NativeBugsnagPerformance!.discardNativeSpan).not.toHaveBeenCalled()
+      expect(NativeBugsnagPerformance!.endNativeSpan).not.toHaveBeenCalled()
+    })
```
gingerbenw, line 194: what's calling `discardSpan` in this test? I assume it's the backgrounding in the previous test

packages/core/lib/span-factory.ts
```diff
@@ -159,9 +168,15 @@ export class SpanFactory<C extends Configuration> {
     }
 
     this.spanAttributesSource.requestAttributes(span)
+    this.sendForProcessing(span, endTime)
+  }
 
-    const spanEnded = span.end(endTime, this.sampler.spanProbability)
+  protected discardSpan (span: SpanInternal, endTime: number) {
+    span.end(endTime, this.sampler.spanProbability)
+  }
```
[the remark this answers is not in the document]
yousif-bugsnag: yeah, I guess it could - the end time isn't public and the `SpanEnded` isn't used, so it's really just there to mark the span as invalid.

packages/platforms/react-native/lib/platform-extensions.tsx
```diff
@@ -31,15 +31,16 @@ export const platformExtensions = (appStartTime: number, clock: Clock, schema: R
   },
   attach: function (config?: ReactNativeAttachConfiguration) {
     const logger = schema.logger.validate(config?.logger) ? config.logger : schema.logger.defaultValue
-    const isNativePerformanceAvailable = NativeBugsnagPerformance !== null && NativeBugsnagPerformance.isNativePerformanceAvailable()
+    const platform = Platform.OS ? 'Cocoa' : 'Android'
```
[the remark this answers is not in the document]
yousif-bugsnag: doh!

packages/platforms/react-native/tests/span-factory.test.ts
yousif-bugsnag: yeah, that's a mistake - good spot!

yousif-bugsnag: I'll update both tests to use `DISCARD_END_TIME` to make things clearer

PR #558, opened 2025-01-08
Turbomodule changed to never be null, default object used instead

DariaKunoichi:
# Goal
Remove check for null in turbomodules loading.

## Changeset
Added a default proxy for situations where neither new nor old turbomodule can be loaded.

## Testing
Changed unit tests for id generation - by default tests always fallback to JS implementation because proxy's return value is empty. But no null check needed.

### DIFF
--- a/packages/platforms/react-native/lib/native.ts
+++ b/packages/platforms/react-native/lib/native.ts
@@ -7,32 +7,22 @@ declare const global: {
 
 const isTurboModuleEnabled = () => global.__turboModuleProxy != null
 
-const LINKING_ERROR =
-  'The package \'BugsnagReactNativePerformance\' doesn\'t seem to be linked.'
-
-export const NativeBsgModule = isTurboModuleEnabled()
+const NativeBsgModule = isTurboModuleEnabled()
   ? TurboModuleRegistry.get('BugsnagReactNativePerformance')
   : NativeModules.BugsnagReactNativePerformance
 
-const NativeBugsnagPerformance = NativeBsgModule || new Proxy(
-  {
-    getDeviceInfo: () => undefined,
-    requestEntropy: () => '',
-    requestEntropyAsync: async () => '',
-    getNativeConstants: () => ({ CacheDir: '', DocumentDir: '' }),
-    exists: async (path: string) => false,
-    isDir: async (path: string) => false,
-    ls: async (path: string) => [],
-    mkdir: async (path: string) => '',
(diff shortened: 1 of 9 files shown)


packages/platforms/react-native/lib/native.ts
```diff
@@ -7,8 +7,32 @@ declare const global: {
 
 const isTurboModuleEnabled = () => global.__turboModuleProxy != null
 
-export const NativeBugsnagPerformance = isTurboModuleEnabled()
+const LINKING_ERROR =
+  'The package \'BugsnagReactNativePerformance\' doesn\'t seem to be linked.'
+
+export const NativeBsgModule = isTurboModuleEnabled()
```
yousif-bugsnag, line 13: I don't think we need to export this as it's not used outside of this file

packages/platforms/react-native/lib/native.ts
```diff
@@ -7,8 +7,32 @@ declare const global: {
...
+    readFile: async (path: string, encoding: string) => '',
+    unlink: async (path: string) => { },
+    writeFile: async (path: string, data: string, encoding: string) => { }
+  },
+  {
+    get () {
+      throw new Error(LINKING_ERROR)
+    }
+  }
+)
```
yousif-bugsnag, line 36: At the moment this will never actually use the fallback implementation and the proxy will just throw an error if any of the methods are called (that's what the get handler does)

I don't think we need to use a proxy here - we can just return the object directly if the native module is null. 

```suggestion
const NativeBugsnagPerformance = NativeBsgModule || {
  getDeviceInfo: () => undefined,
  requestEntropy: () => '',
  requestEntropyAsync: async () => '',
  getNativeConstants: () => ({ CacheDir: '', DocumentDir: '' }),
  exists: async (path: string) => false,
  isDir: async (path: string) => false,
  ls: async (path: string) => [],
  mkdir: async (path: string) => '',
  readFile: async (path: string, encoding: string) => '',
  unlink: async (path: string) => { },
  writeFile: async (path: string, data: string, encoding: string) => { }
}
```

(you might also need to update the Spec interface in `NativeBugsnagPerformance.ts` because `getDeviceInfo` might return undefined)

yousif-bugsnag: I think it would be good to have a unit test for this as well

yousif-bugsnag (approved):

PR #559, opened 2025-01-08
Do not delegate automatic navigation spans to native SDK

yousif-bugsnag:
## Goal

Since we can't get accurate rendering metrics for automatic navigation spans (i.e. those created by the navigation library plugins) because they can only be created after the navigation has occurred, we should not be delegating these spans to the native SDK in the first place.

This PR adds support for internally excluding certain first-class spans from the native integration, and ensures that only manual navigation spans are delegated to the native SDK.

## Design

- Moved the navigation span logic into `ReactNativeSpanFactory`
- Added a new internal `ReactNativeSpanOptions` interface to allow callers to disable the native span functionality for individual spans.

## Testing

Updated unit tests


gingerbenw (approved):
Neat! 👌🏻

PR #560, opened 2025-01-09
Periodically clean up native spans that are open for > 1 hour

yousif-bugsnag:
## Goal

The Turbo Module holds references to the native spans it creates so that they can be retrieved and ended later. This has the potential for memory leaks if for some reason the spans are never ended in JS. 

This PR tries to mitigate this by periodically discarding any native spans that have been open for a long time (> 1 hour)

## Design

When attach is called in JS, the Turbo Module schedules a task to run every hour and discard any spans with a start time more than 1 hour in the past.

## Testing

Tested manually as it's not possible to automate

### DIFF
--- a/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
+++ b/packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
@@ -3,6 +3,7 @@ package com.bugsnag.reactnative.performance;
 import android.annotation.SuppressLint;
 import android.content.pm.PackageInfo;
 import android.os.Build;
+import android.os.SystemClock;
 
 import androidx.annotation.Nullable;
 
@@ -12,7 +13,11 @@ import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import java.security.SecureRandom;
-import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import com.bugsnag.android.performance.BugsnagPerformance;
 import com.bugsnag.android.performance.SpanOptions;
@@ -34,6 +39,8 @@ class NativeBugsnagPerformanceImpl {
 
   private boolean isNativePerformanceAvailable = false;
 
+  private boolean spanCleanupTaskScheduled = false;
+
   /**
    * A map of open native spans, keyed by the span ID and trace ID,
    * so that they can be retrieved and closed/discarded from JS.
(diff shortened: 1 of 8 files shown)


packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -240,6 +253,16 @@ private SpanOptions readableMapToSpanOptions(ReadableMap jsOptions) {
...
 
+  private void discardLongRunningSpans() {
+    long oneHourAgo = SystemClock.elapsedRealtimeNanos() - TimeUnit.HOURS.toNanos(1);
+    for (Map.Entry<String, SpanImpl> entry : openSpans.entrySet()) {
+      SpanImpl span = entry.getValue();
+      if (span.getStartTime$internal() < oneHourAgo) {
+        openSpans.remove(entry.getKey());
+        span.discard();
+      }
+    }
```
lemnik, line 264: nit: I think an `Iterator.remove` is preferable to `Map.remove` here

```suggestion
    Iterator<Map.Entry<String, SpanImpl>> entries = openSpans.entrySet();
    while (entries.hasNext()) {
      Map.Entry<String, SpanImpl> entry = entries.next();
      SpanImpl span = entry.getValue();
      if (span.getStartTime$internal() < oneHourAgo) {
        entries.remove();
        span.discard();
      }
    }
```

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
```diff
@@ -119,6 +126,12 @@ public WritableMap getNativeConfiguration() {
       return null;
     }
 
+    if (!spanCleanupTaskScheduled) {
+      ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+      scheduler.scheduleWithFixedDelay(this::discardLongRunningSpans, 1, 1, TimeUnit.HOURS);
```
lemnik, line 131: Could this perhaps be replaced with a main-thread `Handler` so that we don't require a whole cleanup thread?

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
yousif-bugsnag: Updated 👍🏾

packages/platforms/react-native/android/src/main/java/com/bugsnag/android/performance/NativeBugsnagPerformanceImpl.java
yousif-bugsnag: Updated 👍🏾

packages/platforms/react-native/ios/BugsnagReactNativePerformance.mm
```diff
@@ -216,35 +231,40 @@ static uint64_t hexStringToUInt64(NSString *hexString) {
...
+            // We need to reinstate the bugsnag.sampling.p attribute here as it might not be re-populated on span end
+            nativeSpan.attributes[@"bugsnag.sampling.p"] = @(nativeSpan.samplingProbability);
+            
+            // If the end time is later than the current end time, update it
+            NSDate *nativeEndTime = [NSDate dateWithTimeIntervalSince1970: endTime / NSEC_PER_SEC];
+            if ([nativeEndTime timeIntervalSinceDate:nativeSpan.endTime] > 0) {
+                [nativeSpan markEndTime:nativeEndTime];
+            }
+            
+            [nativeSpan sendForProcessing];
```
kstenerud, line 252: I think that these parts don't need to be inside of the synchronized code block?

yousif-bugsnag: ah yeah, good point 👍🏾

lemnik (approved):
LGTM

PR #574, opened 2025-01-23
Release v2.11.0

yousif-bugsnag:
### Added

- (react-native) Added `BugsnagPerformance.attach` method to support integration with native BugSnag Performance Monitoring SDKs for Android and iOS [#573](https://github.com/bugsnag/bugsnag-js-performance/pull/573)

### Changed

- (react-native) Replaced `react-native-file-access` peer dependency with internal native module implementation. This peer dependency is no longer required by the SDK. [#561](https://github.com/bugsnag/bugsnag-js-performance/pull/561)


gingerbenw (approved):
Happy from my side - but would be happy to await a changelog review from @tomlongridge

tomlongridge (approved):

PR #578, opened 2025-02-19
Support React Native 0.77

yousif-bugsnag:
## Goal

Adds support for React Native 0.77.

## Design

- This required a small fix to the logic that detects whether turbo modules are enabled - previously this relied on the presence of `global.__turboModuleProxy` but this is no longer set in RN 0.77 bridgeless mode (which is the default).

 - 0.77 tests have been added to CI and the test matrices have been updated to test against the following versions:
	- Latest 3 RN versions
	- Older even versions of RN, going back to 0.72
	- 0.64

- Test fixture generation script and pipelines have been updated to cope with New Architecture being enabled by default in 0.76 - `RCT_NEW_ARCH_ENABLED` must now always be set when generating test fixtures. Also made some general improvements to the script to eliminate the need for some replacement files

## Testing

Covered by a full CI run


.buildkite/react-native-navigation-pipeline.full.yml
```diff
@@ -90,32 +92,34 @@ steps:
...
-          - ./bin/generate-react-native-fixture
-        matrix:
-          setup:
-            rn_version:
-              - "0.71"
-        retry:
-          automatic:
-            - exit_status: "*"
-              limit: 1
+      # Skipped due to issues with react-native-navigation when new arch is enabled
```
yousif-bugsnag, line 95: Up until now these tests were actually running against the old architecture because we were setting `RCT_NEW_ARCH_ENABLED` to `true` instead of `1`

.buildkite/react-native-navigation-pipeline.full.yml
```diff
@@ -244,43 +248,45 @@ steps:
...
-        concurrency_group: "bitbar"
-        concurrency_method: eager
-        matrix:
-          setup:
-            ios_version:
-              - "14"
-            rn_version:
-              - "0.71"
+      # Skipped due to issues with react-native-navigation when new arch is enabled
+      # TODO: Investigate and try to re-enable when we add tests for more recent React Native versions
```
gingerbenw, line 252: can we do this with a `skip:` rather than commenting it out?

gingerbenw: I may be confusing some other yaml config here, doesn't look like we can.

.buildkite/react-native-navigation-pipeline.full.yml
yousif-bugsnag: The environment variable is used by React Native to enable/disable new arch for iOS builds, and it doesn't accept `true`, which is why the script now enforces `1`/`0`

gingerbenw (approved):

PR #580, opened 2025-02-20
Update permissions for github actions

gingerbenw:
## Goal

Ensure github actions have appropriate for each job

### DIFF
--- a/.github/workflows/compare-pr-stats.yml
+++ b/.github/workflows/compare-pr-stats.yml
@@ -2,8 +2,7 @@ name: Compare PR stats
 
 on: [pull_request]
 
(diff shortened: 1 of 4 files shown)


.github/workflows/compare-pr-stats.yml
```diff
@@ -2,6 +2,9 @@ name: Compare PR stats
 
 on: [pull_request]
 
+permissions:
+  pull-requests: write
```
tomlongridge, line 6: We should have read-all at the top level and then give each job the write permissions that it needs.

tomlongridge (approved):

PR #585, opened 2025-02-24
Allow pull request write permissions to Compare PR stats action

gingerbenw:
## Goal

Allow pull request write permissions to Compare PR stats action

### DIFF
--- a/.github/workflows/compare-pr-stats.yml
+++ b/.github/workflows/compare-pr-stats.yml
@@ -12,19 +12,23 @@ concurrency:
 
 jobs:
   base-branch:
+    permissions:
+      pull-requests: write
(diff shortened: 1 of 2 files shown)


.github/workflows/compare-pr-stats.yml
```diff
@@ -2,7 +2,8 @@ name: Compare PR stats
 
 on: [pull_request]
 
-permissions: read-all
+permissions:
+  pull-requests: write
```
gingerbenw, line 6: @tomlongridge these are the same permissions we are currently using on the notifier, the previous configuration was invalid

tomlongridge (approved):

PR #594, opened 2025-03-17
End to end tests

AnastasiiaSvietlova:
## Goal

Add end to end tests for component lifecycle spans in react

## Changeset

This PR also resolves a bug in component span instrumentation where component spans were not ended if the app state was not 'ready'

## Testing

Added new end to end tests in React


packages/plugin-react-performance/lib/with-instrumented-component.tsx
```diff
@@ -30,7 +30,7 @@ class InstrumentedComponent extends React.Component<InstrumentedComponentProps>
   }
 
   async componentDidMount () {
-    if (this.componentMountSpan && BugsnagPerformance.appState === 'ready') {
```
gingerbenw, line 33: @AnastasiiaSvietlova this logic was the wrong way around - the fact we hadn't detected this probably means the unit tests aren't sufficient

lemnik (approved):

yousif-bugsnag (approved):

PR #609, opened 2025-04-10
Update react native example app

yousif-bugsnag:
## Goal

Update the RN example app to 0.78 and modify the setup so that bugsnag packages are referenced directly from source.

This should improve the developer experience a lot as it makes it easier to use Xcode/Android Studio during development via the example app project.

## Design

App contents are mostly copied from the old rn072 fixture. Had to make some tweaks to the metro configuration to get it to resolve the symlinked packages from the repo.

## Testing

Tested manually


gingerbenw (approved):
looks good to me! glad that metro config worked :+1:

PR #615, opened 2025-04-23
Stop using legacy Selenium driver 

twometresteve:
## Goal

Stop using the legacy Selenium driver, it's not needed.

## Changeset

As part of working this change, I discovered a bug in Maze Runner that meant we hadn't be running on Edge 80.  Having fixed that, I discovered that the navigation scenarios fail - these are now skipped pending further investigation.

## Testing

Covered by unblocking all browser tests on CI.

### DIFF
--- a/test/browser/features/navigation-changes.feature
+++ b/test/browser/features/navigation-changes.feature
@@ -1,3 +1,5 @@
+# Skipped on Edge 80 pending PLAT-14141
+@skip_edge_80
 @requires_fetch_keepalive
 Feature: Navigation changes
     @skip_span_time_validation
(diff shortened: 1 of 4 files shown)


test/browser/features/navigation-changes.feature
```diff
@@ -1,3 +1,5 @@
+# Skipped on Edge 80 pending PLAT-14141
+@skip_edge_80
```
twometresteve, line 2: This work uncovered a bug in Maze Runner that meant we weren't actually running tests on Edge 80 at all.  Having resolved that I found these scenerios are failing and need further investigation.

joshedney (approved):

gingerbenw:
Edge 80 tests have been covered in [this PR](https://github.com/bugsnag/bugsnag-js-performance/pull/675) - there appears to be a bug in this version of Edge with keepalive cross origin requests

PR #620, opened 2025-05-12
Allow easy cross-layer span parenting

yousif-bugsnag:


packages/core/lib/remote-parent-context.ts
```diff
@@ -0,0 +1,43 @@
...
+import { isParentContext, isSpanContext } from './validation'
+
+const TRACE_PERENT_REGEX = /^00-([0-9a-f]{32})-([0-9a-f]{16})-[0-9]{2}$/
+
+export default class RemoteParentContext implements ParentContext {
+  readonly id: string
+  readonly traceId: string
+
+  constructor (maybeParentContext: ParentContext | string, traceId?: string) {
+    if (isParentContext(maybeParentContext)) {
```
yousif-bugsnag, line 12: Is this constructor confusing? Would it make more sense to only support `ParentContext` objects here?

```
new RemoteParentContext({ id: 'span-id', traceId: 'trace-id' })
```

lemnik: Yes, I think that might be significantly clearer.

packages/core/lib/remote-parent-context.ts
```diff
@@ -0,0 +1,43 @@
...
+    return RemoteParentContext.toTraceParentString(this)
+  }
+
+  static parseTraceParent (encodedString: string) {
+    if (!TRACE_PERENT_REGEX.test(encodedString)) {
+      return undefined
+    }
+
+    const [, traceId, spanId] = encodedString.split('-')
+    return new RemoteParentContext(spanId, traceId)
```
lemnik, line 31: nit: use the captured groups in the regex instead of splitting the string

```suggestion
    const result = TRACE_PERENT_REGEX.exec(encodedString)
    if (!result) {
      return undefined
    }

    const [, traceId, spanId] = result
    return new RemoteParentContext(spanId, traceId)
```

lemnik (approved):

PR #626, opened 2025-05-16
(plugin-react-navigation): Add registerNavigationContainerRef for Expo Router support

yousif-bugsnag:
## Goal

Rework the react-navigation plugin to support registering a navigation container ref in addition to the existing `createNavigationContainer`.

As part of this, the plugin has been refactored to replace the `NavigationContext` and its provider with a new `NavigationTracker` class for managing navigation spans and state.

This is primarily so that we can support Expo Router where the `NavigationContainer` component is not directly exposed, but also provides some additional improvements:

- Support for the static configuration API (introduced in React Navigation v7) - as with Expo Router, this doesn't use the NavigationContainer component directly.

- Navigation start times are now recorded at the start of the navigation action (using the `__unsafe_action__` event) - this means navigation spans have a much more accurate and meaningful start time and duration.

This PR also adds e2e tests for Expo Router.

## Design

- Removed `NavigationContext` and associated provider and replaced with a new `NavigationTracker` class to handle navigation spans and app state. This includes methods for blocking, unblocking, and ending navigation spans.

- The `CompleteNavigation` component now interacts with the `NavigationTracker` through the Bugsnag plugin instead of relying on `NavigationContext` - this allows it to work as before without us having to wrap the `NavigationContainer` with a context provider.

- The `createNavigationContainer` function now uses `NavigationTracker` internally instead of `NavigationContextProvider`.
- 

## Testing

- Refactored and updated unit tests
- Added e2e tests for Expo Router


packages/plugin-react-navigation/test/navigation-tracker.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    const setAppState = jest.fn((state: AppState) => { appState = state })
+    const spanFactory = new MockReactNativeSpanFactory()
+
+    const mockNavigationContainer = mockNavigationContainerRef()
+    const navigationTracker = new NavigationTracker(spanFactory as unknown as ReactNativeSpanFactory, setAppState)
+    navigationTracker.configure(mockNavigationContainer as unknown as NavigationContainerRef<ReactNavigation.RootParamList>)
+
+    expect(appState).toBe('starting')
+
+    // TODO: Initial route should not create a span??
```
gingerbenw, line 56: I'm guessing this got copied over from older code? worth noting it - I don't think the first route _should_ count as a navigation, but just raising my hand about this one

gingerbenw (approved):

yousif-bugsnag:
force-merging due to issue with CodeQL

PR #629, opened 2025-05-20
Replace Docker with native macOS fixture builds for browser tests

twometresteve:
## Goal

Build browser test fixtures using macOS server instead of Docker images.

## Design

We had an issue with the use of Docker for building the and running the browser tests.  What was supposed to be a CDN build was randomly coming through as NPM and vice versa.

I've sidestepped the issue completely by using our macOS servers to build the test fixtures and using the pre-built Maze Runner image directly.  The end result is a CI build that's quicker to run by a few minutes and there is a little less to maintain in the files.

## Changeset

- The custom Docker build steps are removed, replaced with macOS build steps for the fixture files.
- I changed the `USE_CDN_BUILD` and `EXTRA_STEP_LABEL` variables to just use `BUILD_MODE` (with values of `NPM` or `CDN`) - to be explicit about the mode and reduce the number of moving parts.
- When `pipeline-trigger.js` uploads pipeline files it now needs to set the `BUILD_MODE` environment variable, so I've added a generic mechanism to the existing setup to allow us to set an arbitrary environment.

## Testing

Covered by CI, unblocking all the browser tests and paying careful attention to the step titles and Maze Runner output for CDN and NPM builds.  I've also conducted a number of tests exercises the various various ways that pipeline files can get triggered:
- Automatically based on PR changes
- Automatically by commit message
- Manually triggered


.buildkite/browser-pipeline.yml
```diff
@@ -73,10 +70,18 @@ steps:
...
         timeout_in_minutes: 30
         plugins:
+          artifacts#v1.9.4:
+            download:
+              - "fixture-packages-${BUILD_MODE}.zip"
+              - from: "test/browser/features/fixtures/.bugsnag-browser-version-${BUILD_MODE}"
+                to: "test/browser/features/fixtures/.bugsnag-browser-version"
+            upload: "./test/browser/maze_output/failed/**/*"
+          improbable-eng/metahook:
+            pre-command: "unzip -o fixture-packages-${BUILD_MODE}.zip"
```
twometresteve, line 84: This is a slightly annoying dance we have to do, due to the fact that the artifacts plugin has poor support for compressed artifacts (the `compressed` attribute is shard between the upload and download).

.buildkite/browser-pipeline.yml
```diff
@@ -4,45 +4,45 @@ agents:
...
-              build: browser-maze-runner-bs
-              image-repository: 855461928731.dkr.ecr.us-west-1.amazonaws.com/js
-              cache-from: browser-maze-runner-bs:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:performance-ci-browser-${BRANCH_NAME}-${USE_CDN_BUILD}
-          - docker-compose#v4.12.0:
-              push: browser-maze-runner-bs:855461928731.dkr.ecr.us-west-1.amazonaws.com/js:performance-ci-browser-${BRANCH_NAME}-${USE_CDN_BUILD}
+        agents:
+          queue: macos-14
         env:
-          USE_CDN_BUILD: "${USE_CDN_BUILD}"
+          BUILD_MODE: "${BUILD_MODE}"
```
twometresteve, line 14: There's a problem here when the pipeline is triggered based on a file change.  I need to have that script set the environment variable when uploading the pipeline file.

PR #630, opened 2025-05-21
Fix release script

yousif-bugsnag:
## Goal

Fixes an issue with the release script where the `lerna version` command was causing unwanted changes to the `package-lock.json`

## Design

the `lerna version` command runs npm install with `legacy_peer_deps` internally, which result sin changes to the lockfile, so we explicitly set this to false to work around the issue.


gingerbenw (approved):

yousif-bugsnag:
force-merging due to issue with CodeQL

PR #632, opened 2025-05-28
Plat 14191/svelte kit navigation

gingerbenw:

### DIFF
--- a/packages/svelte-kit/lib/svelte-kit-routing-provider.ts
+++ b/packages/svelte-kit/lib/svelte-kit-routing-provider.ts
@@ -1,0 +1,35 @@
+import type { RouteChangeSpan, RoutingProvider, StartRouteChangeCallback } from '@bugsnag/browser-performance'
+import { onSettle } from '@bugsnag/browser-performance'
+import type { AfterNavigate, BeforeNavigate } from '@sveltejs/kit'
+
+type NavigateCallback<T> = (callback: (navigation: T) => void) => void
+
+export class SvelteKitRoutingProvider implements RoutingProvider {
+  constructor (private beforeNavigate: NavigateCallback<BeforeNavigate>, private afterNavigate: NavigateCallback<AfterNavigate>) {}
+
+  // SvelteKit exposes the route as a URL in the beforeNavigate hook
+  // so we can use it directly by converting to a string.
+  resolveRoute (url: URL) {
+    return url.toString()
+  }
+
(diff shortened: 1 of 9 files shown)


packages/svelte-kit/lib/svelte-kit-routing-provider.ts
```diff
@@ -0,0 +1,35 @@
...
+type NavigateCallback<T> = (callback: (navigation: T) => void) => void
+
+export class SvelteKitRoutingProvider implements RoutingProvider {
+  constructor (private beforeNavigate: NavigateCallback<BeforeNavigate>, private afterNavigate: NavigateCallback<AfterNavigate>, private base = window.location.origin) {}
+
+  // SvelteKit exposes the route as a URL in the beforeNavigate hook
+  // so we can use it directly by converting to a string.
+  resolveRoute (url: URL) {
+    return url.toString()
+  }
```
yousif-bugsnag, line 14: I'm a bit confused by this. I think the route change plugin expects `resolveRoute` to convert an absolute url to a route. 

When the plugin is initialised, it calls `resolveRoute` with the current URL (`window.location.href`) to get the initial route. So I think this would set that as the full URL, which I don't think is what we want.

It also uses the URL to set the `bugsnag.browser.page.url` attribute on the span - so the `bugsnag.browser.page.route` and `bugsnag.browser.page.url` attributes would always be the same?

test/browser/features/svelte-kit.feature
```diff
@@ -0,0 +1,30 @@
...
+        And I click the element "profile"
+        And I wait to receive at least 2 spans
+
+        Then a span named "[RouteChange]/contact/[contactId]" contains the attributes: 
+            | attribute                                 | type         | value                          | 
+            | bugsnag.span.category                     | stringValue  | route_change                   |
+            | bugsnag.browser.page.route                | stringValue  | /contact/[contactId]           |
+            | bugsnag.browser.page.previous_route       | stringValue  | /                              |
+            | bugsnag.browser.page.route_change.trigger | stringValue  | link                           |
+            # | bugsnag.browser.page.url                  | stringValue  | /contact/1                    |
```
yousif-bugsnag, line 18: should these be commented out?

yousif-bugsnag (approved):

yousif-bugsnag (dismissed):

gingerbenw:
Sorry @yousif-bugsnag - last one, I promise!

PR #638, opened 2025-06-09
Updated plugin interface and PluginContext

yousif-bugsnag:
## Goal

This PR reworks the existing plugin interface to allow plugins to register span callbacks and span control providers.

## Design

- Introduced `PluginContext` to allow plugins to register span callbacks and span control providers with the client, and `PluginManager` to manage the installation, initialization and retrieval of plugins.

- Refactored the plugin interface, replacing `configure` with separate `install` and `start` methods.

- `SpanFactory` is no longer directly exposed to plugins via `configure`, and external plugins that previously relied on `SpanFactory` for starting and ending spans now do so through the public APIs (see `plugin-react-navigation` and `plugin-react-native-navigation`).

- App state handling has been centralized into a separate module, removing the need to pass the getters and setters around.

## Testing

Added and updated unit tests for new plugin structure.


yousif-bugsnag:
> Do you think it would be useful to log a message in the plugins if we try to start and it's not enabled? I don't see when that would happen - but just a thought.

I don't think we really want to do that  since all of these are internal plugins that are always installed but can be disabled via config - so we'd be logging messages whenever the config option is turned off.

External plugins (those added in config) can't be disabled like that - they are either present in the configuration or they're not.

PR #643, opened 2025-06-16
Set default endpoints based on API key

richardelms:
## Goal

Set default endpoints based on API key. If no endpoint is configured, payloads should be sent to `*.otlp.insighthub.smartbear.com` if the API key starts with `00000`.

## Testing

Added new unit tests


yousif-bugsnag (approved):
LGTM

PR #645, opened 2025-06-19
iOS Remote Span controls

lemnik:
## Goal
Allow cross-layer control of native spans from ReactNative on iOS.

## Design
Introduce an iOS version of the Android `BugsnagRemoteSpans` module and `BugsnagRemoteSpansPlugin` that can be used by ReactNative JS code to retrieve and make changes to spans created in the native layer.

## Changeset
This PR also includes an update to the Android module to make it old architecture compatible.

## Testing
Enabled the remote-spans end2end tests for iOS & old architecture


packages/plugin-react-native-remote-spans/android/src/main/java/com/bugsnag/reactnative/performance/remotespans/BugsnagRemoteSpans.java
```diff
@@ -36,12 +35,14 @@ public class BugsnagRemoteSpans extends NativeBugsnagRemoteSpansSpec {
...
     public BugsnagRemoteSpans(ReactApplicationContext reactContext) {
-        super(reactContext);
+        this.reactContext = reactContext;
     }
 
-    @Override
-    public @Nullable WritableMap getSpanIdByName(String spanName) {
+    @Nullable
+    public WritableMap getSpanIdByName(String spanName) {
         NativeSpanAccessPlugin nativeSpanAccessPlugin = NativeSpanAccessPlugin.getInstance();
```
yousif-bugsnag, line 46: We should probably rename this to `BugsnagRemoteSpansPlugin` for consistency with iOS and JS

lemnik: Now that you say that, I think maybe the `Remote` plugins should be renamed - since they could imply client/server relationship rather than a clear native/js relationship?

.buildkite/scripts/packages.json
```diff
@@ -49,7 +50,8 @@
             "packages/platforms/react-native",
             "packages/delivery-fetch",
             "packages/request-tracker",
-            "packages/plugin-react-native-navigation"
+            "packages/plugin-react-native-navigation",
+            "packages/plugin-react-native-remote-spans"
```
yousif-bugsnag, line 54:
```suggestion
            "packages/plugin-react-native-span-access"
```

yousif-bugsnag (approved):

PR #662, opened 2025-07-16
Fix iOS crash from requestEntropyAsync

maxphillipsdev:
## Goal

When calling `startSpan` the `requestEntropy` function is called once to generate values for the span IDs. After the first 1000 spans the values are regenerated using the `requestEntropyAsync` function but the objc file has a typo which causes it to crash when called.

**Bug sample:** 
[link](https://github.com/maxphillipsdev/bugsnag-js-performance/tree/startSpan-crash-repro)

**Sample instructions:** 
I lowered the `CALLS_BEFORE_POOL_REFRESH ` value from 1000 to 10. 
Just click send on "Navigation Span" 10 times and look in the console to see each time the id generator function is called. After 10 times it will crash from a missing selector.

image

## Changeset

Just fixed the typo in the argument name.

## Testing

Ran unit tests and manually tested in sample app.


yousif-bugsnag:
Closing this in favour of https://github.com/bugsnag/bugsnag-js-performance/pull/667

PR #672, opened 2025-07-22
ReactNative: avoid representing unix nanoTime as a double

lemnik:
## Goal
Avoid precision loss when dealing with the unix timestamp in ReactNative.

## Design
Rework the ReactNative `Clock.toUnixTimestampNanoseconds` to use two `number` values representing 32bit integers to form two halves of an `int64` and convert directly from the format into a string.

Whenever we transfer a timestamp to or from the native layer, the code now uses a `string` timestamp in the wire format to avoid any loss of precision.

The iOS code also now uses an improved conversion to go from a string representing an `int64` to an `NSNumber` representing the standard iOS unix timestamp and then to an `NSDate` for compatibility with `bugsnag-cocoa-performance`.

## Testing
Updated the unit tests for the new format, along with more reliable runs of the existing end2end tests.


packages/plugin-react-native-span-access/lib/index.ts
yousif-bugsnag, line 1: FYI this code has moved to a different file in the integration branch - it might be easier to just target that branch for these changes instead of next, since we'll be merging it all down after this anyway

yousif-bugsnag (approved):

gingerbenw (approved):
Can't quite get my head around some of the maths in `lib/clock.ts` but the logic is good and I'm happy that the tests cover it

PR #675, opened 2025-07-28
Validate skipped tests on Edge 80

gingerbenw:
## Goal

Update `requires_fetch_keepalive` definition to start from Edge 81 as Edge 80 has a bug where cross-origin preflight requests are not supported with `keepalive`

Screenshot 2025-07-31 at 12 33 12

## Testing

Covered by existing ci (edge 80 tests are still skipped but will run again when we update the pipeline to use 81)

### DIFF
--- a/test/browser/features/fixtures/packages/navigation-changes/index.html
+++ b/test/browser/features/fixtures/packages/navigation-changes/index.html
@@ -6,13 +6,13 @@
     </head>
     <body>
         <h1>navigation-changes</h1>
-        
+
         <button id="add-span-to-batch">add-span-to-batch</button>
-        
+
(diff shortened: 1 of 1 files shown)


test/browser/features/lib/browser.rb
```diff
@@ -36,8 +36,8 @@ def supports_fetch_keepalive?
       @version >= 66
 
     when "edge"
-      # support added in Edge 15
-      @version >= 15
+      # On Edge 80 keepalive doesn't work for cross-origin preflight requests 
+      @version >= 81
```
gingerbenw, line 40: I wonder if this should still be 15 and up but exclude this specific version, as this definition should really cover when this feature is actually supported?

PR #679, opened 2025-08-04
Calling setAttribute with null/undefined should clear the attribute

yousif-bugsnag:
## Goal

Previously calls to `setAttribute` with `null` or `undefined` values were ignored, so there was no way to remove span attributes. This PR updates the implementation so that calling `setAttribute` with a `null` or `undefined` attribute value clears the attribute from the span.

## Design

Updated the `SpanAttribute` type to accept null/undefined values and updated the `set` and `setCustom` methods to remove the attribute if the value is `null` or `undefined`

## Testing

Updated the attribute unit tests and updated the maze runner scenario to verify that span attributes can now be removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,12 @@
 ### Changed
 
 - Reduce batch time in development environment [#673](https://github.com/bugsnag/bugsnag-js-performance/pull/673)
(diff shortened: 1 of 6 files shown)


lemnik (approved):
LGTM - With one small type suggestion

lemnik (approved):
LGTM

PR #699, opened 2025-09-03
Add retries to Appium steps

joshedney:
## Goal

Add automatic retries to Appium steps for the exit code 103

## Testing

Covered by CI


twometresteve (approved):
There's a couple of commented out steps that the retry could also be added to, but this is fine as it is.

PR #712, opened 2025-09-30
Native test utilities for React Native

yousif-bugsnag:
## Goal

Currently the React Native tests are driven primarily from the React Native side, with a Turbo Module for native operations.

This has worked okay up until now, but as we introduce more native functionality for React Native we need more control over the native side of the test fixtures for things like starting the native SDKs.

This PR introduces new native `BugsnagTestUtils` modules for Android and iOS to handle some of the native operations that were previously handled by the scenario launcher Turbo Module.

## Design

Initially the operations moved to the new utility modules are:
- reading/writing startup config
- starting the native performance SDKs.

This will make it easier to start the native SDKs independently of the Turbo Module so that we can test things like native app starts effectively.

The new native modules are installed as native dependencies of the test fixture and are referenced by the turbo module

## Changeset

- Added new BugsnagTestUtils modules for Android and iOS
- Updated the fixture generation scripts to install the new utility modules into dynamic fixtures
- Moved the scenario launcher up a few directory levels into `test/react-native`

## Testing

Covered by CI


test/react-native/features/fixtures/expo/withNativeTestUtilsAndroid.js
```diff
@@ -0,0 +1,40 @@
...
+  // Modify settings.gradle to include the native test utils module
+  config = withSettingsGradle(config, (config) => {
+    const settingsGradleContent = config.modResults.contents;
+    const includeDependency = 'include("bugsnag-test-utils")';
+
+    // assumes native-test-utils has been copied into the fixture directory
+    const dependencyPath = 'project(":bugsnag-test-utils").projectDir = file("../native-test-utils/android")';
+    config.modResults.contents = `${settingsGradleContent}\n${includeDependency}\n${dependencyPath}`;
+    // Check if the module is already included to avoid duplicates
+    if (!settingsGradleContent.includes('bugsnag-test-utils')) {
```
gingerbenw, line 13: isn't this exactly the same as the operation above?

yousif-bugsnag: doh, yes it is - good spot. It's harmless because it only gets run once anyway, but I've removed the duplicate line 👍🏾

gingerbenw (approved):
Seems sensible to me, but I'd appreciate another pair of eyes on the native code as I'm sure there's something there I don't fully get!

PR #730, opened 2025-10-21
React Native client factory

hamudi-ahmed:
## Goal

Add support for creating specialized React Native clients via a new exported method `createReactNativeClient`. This allows upstream libraries to create 'wrapper' clients that override or extend specific parts of the default react native client.

## Design

The factory method accepts a `Partial<ClientOptions>` parameter so that specific client options can be overridden with specialized implementations where required, while maintaining the default React Native functionality for others. 

## Changeset

- Added `createReactNativeClient` factory method
- refactored default export to use the new factory method
- added new unit tests and reorganized existing client tests 

## Testing

New unit tests plus existing CI coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- (react-native): Added `createReactNativeClient` factory method for internal use by upstream libraries [#730](https://github.com/bugsnag/bugsnag-js-performance/pull/730)
+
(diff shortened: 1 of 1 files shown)


gingerbenw (approved):
seems good to me - I wonder if we could mock fetch in the react-native setup file, much like XHR?

gingerbenw:
also, this could probably use it's own CHANGELOG entry!

gingerbenw (approved):

PR #735, opened 2025-10-28
Update browser tests to use Maze Runner v10

twometresteve:
## Goal

Update browser tests to use Maze Runner v10.

## Design

This change drove an addition to Maze Runner - the step `I wait to receive a span named {string}`.  Some tests were flaking because extra spans that the scenarios didn't care about were being sent first, leading to the wrong ids being extracted.  I've changed a number of scenarios to use this new step, improving their reliability.  There may be more scenarios that could be migrated to use it too.

## Changeset

I've also made a change to the pipeline to aid debugging and other activities - maze_output/maze_output.zip is also uploaded so that we can inspect the payloads received for passed tests as well as any failed.

## Testing

Covered by CI with both browser triggers released.

### DIFF
--- a/.buildkite/react-native-navigation-pipeline.yml
+++ b/.buildkite/react-native-navigation-pipeline.yml
@@ -217,7 +217,7 @@ steps:
 
       # iOS Test
       # ------------------------------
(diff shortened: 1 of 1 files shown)


yousif-bugsnag (approved):
🚀

DariaKunoichi (approved):

PR #794, opened 2026-03-17
Bugsnag performance SDK support react native updated version 0.84

SB-priyankap:
# Goal

Bugsnag performance SDK support react native updated version 0.84

## Changeset

updated buildkite yml file(support react native latest version and also updated the version of node) and made changes inside android-utils.js for new arch

## Testing

Done local testing as well as buildkite test cases


hamudi-ahmed (approved):
LGTM

PR #806, opened 2026-04-16
fix issue to support Vue router v5

SB-priyankap:
## Goal

fix issue to support Vue router v5

## Changeset

Update vue router to support both v4 and v5

## Testing

Perform unit testing


SB-priyankap:
close PR

PR #825, opened 2026-07-31
Pri/plat sca sprint 19 js performance

SB-priyankap:
## Goal

Fixes for SCA finding

## Changeset

Update packages version 

## Testing

Did UT


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '3.6'
```
SB-jigneshR, line 1: @SB-priyankap : Specific reason to remove version in docker compose ??

SB-priyankap: For fixing license audit issue I have removed this. But without this also it will work. I will fix this and again raise PR

SB-jigneshR (approved):

